Issue #31, opened 2022-11-21
Update SDK to login to remote resources


MichaelClifford:
We need to add a function to allow the user to login to their remote resources through the SDK.

MichaelClifford:
closed by #39

PR #39, opened 2022-11-28
add login, logout functionality for OpenShift cluster 

MichaelClifford:
This PR allows users of the SDK to login and out of an OpenShift cluster as part of their "cluster.up()" call by adding an Authentication object to their cluster config. 

This includes a new file `auth` that contains a number of Authentication classes that can be used depending on the users specific authentication needs.

### DIFF
--- a/src/codeflare_sdk/cluster/auth.py
+++ b/src/codeflare_sdk/cluster/auth.py
@@ -1,0 +1,111 @@
+# Copyright 2022 IBM, Red Hat
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+The auth sub-module contains the definitions for the Authentication objects, which represent
+the methods by which a user can authenticate to their cluster(s). The abstract class, `Authentication`,
+contains two required methods `login()` and `logout()`. Users can use one of the existing concrete classes to
+authenticate to their cluster or add their own custom concrete classes here.  
+"""
+
+import abc
+import openshift as oc
+
+
+class Authentication(metaclass=abc.ABCMeta):
+    """
+    An abstract class that defines the necessary methods for authenticating to a remote environment.
+    Specifically, this class defines the need for a `login()` and a `logout()` function.
+    """
+
+    def login(self):
+        """
+        Method for logging in to a remote cluster.
+        """
+        pass
+
+    def logout(self):
+        """
+        Method for logging out of the remote cluster.
+        """
+        pass
+
+
+class TokenAuthentication(Authentication):
+    """
+    `TokenAuthentication` is a subclass of `Authentication`. It can be used to authenticate to an OpenShift
+    cluster when the user has an API token and the API server address.
+    """
+
+    def __init__(
+        self,
+        token: str = None,
+        server: str = None,
+    ):
+        """
+        Initialize a TokenAuthentication object that requires a value for `token`, the API Token
+        and `server`, the API server address for authenticating to an OpenShift cluster.
+        """
+
+        self.token = token
+        self.server = server
+
+    def login(self):
+        """
+        This function is used to login to an OpenShift cluster using the user's API token and API server address.
+        """
+        token = self.token
+        server = self.server
+        response = oc.invoke("login", [f"--token={token}", f"--server={server}:6443"])
+        return response.out()
+
+    def logout(self):
+        """
+        This function is used to logout of an OpenShift cluster.
+        """
+        response = oc.invoke("logout")
+        return response.out()
+
+
+class PasswordUserAuthentication(Authentication):
+    """
+    `PasswordUserAuthentication` is a subclass of `Authentication`. It can be used to authenticate to an OpenShift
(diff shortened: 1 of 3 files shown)


MichaelClifford:
closes #31

MichaelClifford:
@Maxusmusti @atinsood @erikerlandson 

Re-did the authentication and login set up. Take a look and let me know if this is good to go. Thanks for the initial feedback :)

atinsood:
@MichaelClifford yep, this looks 💯 , 3 questions
- is there a reason why you implemented the `logout` method in the base authentication and not keep it generic. the assumption is that in some cases we might not be using oc and there might be a different way of implementing auth logout
- I think we can make the Authentication class an python abstract class rather than a concrete class https://docs.python.org/3/library/abc.html
- I was thinking if we can make the login/logout operate within context manager  https://docs.python.org/3/library/contextlib.html so rather than doing login/logout as 2 different steps, we can simply do `with authentication()` or something like that. 

but all those are nitpicks, the first pass LGTM

MichaelClifford:
> * is there a reason why you implemented the `logout` method in the base authentication and not keep it generic. the assumption is that in some cases we might not be using oc and there might be a different way of implementing auth logout

This is a good question, that I thought about it a bit and I think gets to a larger question: Would we want to support cases were OpenShift is not the target platform?  If so, we can make it more generic. Otherwise, no matter how you login to the cluster you can always log out the same way, so no need to re-implement.  

> * I think we can make the Authentication class an python abstract class rather than a concrete class https://docs.python.org/3/library/abc.html

Sure, I can make it an abstract class. But what exactly would be the benefit of doing that for our use case?  

> * I was thinking if we can make the login/logout operate within context manager  https://docs.python.org/3/library/contextlib.html so rather than doing login/logout as 2 different steps, we can simply do `with authentication()` or something like that.

Will that work in an interactive notebook environment if your work spans multiple cells? I'll look into it as an option for job submission workflows. I think that would be cool. But I don't think it would work for notebook/interactive workflows (but let me know if im wrong :smile: )

Maxusmusti (changes requested):
Looks great, just needs a rebase (will bring in the format checking), and make to add docstrings/type hints to the new files/classes/functions

atinsood:
> This is a good question, that I thought about it a bit and I think gets to a larger question: Would we want to support cases were OpenShift is not the target platform? If so, we can make it more generic. Otherwise, no matter how you login to the cluster you can always log out the same way, so no need to re-implement.

so I was thinking of a use case when you don't have direct access to the `oc` client but the login/out is happening remotely. or say if the underlying client uses a different binary than `oc`

atinsood:
> Sure, I can make it an abstract class. But what exactly would be the benefit of doing that for our use case?

I think its more of a signal to external contributors going forward that they are supposed to implement their own authentication.

atinsood:
> Will that work in an interactive notebook environment if your work spans multiple cells? I'll look into it as an option for job submission workflows. I think that would be cool. But I don't think it would work for notebook/interactive workflows (but let me know if im wrong 😄 )

I think you should be ok. we are already doing something liek this https://github.com/project-codeflare/codeflare-sdk/pull/39/files#diff-5706a96c5346dd54a22370211260411363c4ed41d798fddf240eb8983b1366eaR99

so I am assuming it will be simialr to that

src/codeflare_sdk/cluster/auth.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(
+        self,
+        token: str = None,
+        server: str = None,
+    ):
+
+        self.token = token
+        self.server = server
+
+    def login(self):
```
Maxusmusti, line 54: Could you also add simple function docstrings? To the init, login, and logout function for each implementation.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -94,6 +94,7 @@ def up(self):
         Applies the AppWrapper yaml, pushing the resource request onto
         the MCAD queue.
         """
+        self.config.auth.login()
```
Maxusmusti, line 97: Do we want to log in when we do cluster.up(), and logout with cluster.down()? Or do we want to wrap each call (up, down, etc) with a login at the beginning, and a logout at the end? The latter is what I thought we were originally going for, but I could see reasons why we may want to do this instead. Curious on your takes @atinsood @MichaelClifford

Maxusmusti (changes requested):
Looks great, no architectural suggestions or anything, just a couple minor nits

src/codeflare_sdk/cluster/auth.py
MichaelClifford: yes, my bad. Should have add those :upside_down_face:

src/codeflare_sdk/cluster/cluster.py
MichaelClifford: In my opinion, I don't think that we want to login and out on every single call. This is not how a user would naturally interact with a cluster, and I'm not sure of the benefits of doing it that way. 

Also, on logout() your token can expire, which would break subsequent login in calls. 

It seems reasonable to me that you would want to login at the start of a sessions (on cluster.up()) and logout at the end (on cluster.down())  

Interested to hear other takes though :smile:

Maxusmusti: Oh the expiration is an excellent point, I hadn't thought about that. I think your implementation is probably the correct way after all

atinsood:
> In my opinion, I don't think that we want to login and out on every single call. This is not how a user would naturally interact with a cluster, and I'm not sure of the benefits of doing it that way.

> Also, on logout() your token can expire, which would break subsequent login in calls.

> It seems reasonable to me that you would want to login at the start of a sessions (on cluster.up()) and logout at the end (on cluster.down())

yeah I think its a fair point. lets go with what we have.

MichaelClifford:
Awesome. Thanks for the feedback all. :smiley:

PR #47, opened 2023-01-13
Add  functions for accessing ray job logs

MichaelClifford:
This PR adds 3 new functions to the Cluster class `list_jobs` , `job_status`, `job_logs` so that a user can get some insight into their submitted ray jobs. 

It also includes a clean up to `cluster_dashboard_uri` to prevent it from returning non-dashboard uri's

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -22,6 +22,7 @@ from os import stat
 from typing import List, Optional, Tuple
 
 import openshift as oc
+from ray.job_submission import JobSubmissionClient
 
 from ..utils import pretty_print
 from ..utils.generate_yaml import generate_appwrapper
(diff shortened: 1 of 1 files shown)


MichaelClifford:
Thanks @Maxusmusti, added the type hints

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -200,6 +203,30 @@ def is_ready(self, print_to_console: bool = True):
...
 
+    def list_jobs(self) -> List:
+        """
+        This method accesses the head ray node in your cluster and lists the running jobs.
+        """
+        dashboard_route = self.cluster_dashboard_uri(namespace=self.config.namespace)
+        client = JobSubmissionClient(dashboard_route)
+        return client.list_jobs()
+
+    def job_status(self, job_id: str):
```
Maxusmusti, line 214: Awesome, I think this line needs a return type ` -> str` but otherwise perfect

Maxusmusti (approved):
LGTM

PR #56, opened 2023-01-27
Adding Image Build/Push to CI

Maxusmusti:
Will build/push new notebook images when `custom-nb-image/VERSION` is updated (on main branch)


.github/workflows/image.yml
```diff
@@ -0,0 +1,37 @@
+# This workflow will build an image for the instascale-controller, then push the image to the `project-codeflare` quay repo
```
MichaelClifford, line 1:
```suggestion
# This workflow will build an image for the codeflare notebook image, then push the image to the `project-codeflare` quay repo
```

MichaelClifford (changes requested):
Small change on the comment. Otherwise, looks great!

.github/workflows/image.yml
```diff
@@ -0,0 +1,37 @@
+# This workflow will build an image for the codeflare notebook image, then push the image to the `project-codeflare` quay repo
+
+name: Image
+
+on:
+  push:
```
MichaelClifford, line 6: Do we want to create a new release on every merged PR? or only when we publish a new release? This looks like it will create a new image with every PR.

Maxusmusti: No this only creates a new image on merged PRs in main that include an update to the "VERSION" file. So when we bump the version for a new release, that is the only time it will build/push

MichaelClifford (approved):

PR #59, opened 2023-02-02
update base notebook image to ODH

MichaelClifford:
Switched the base image to one maintained by ODH community. Updated the version to kick off a new image build. 

I also tested the image on ODH with the mnist and huggingface notebooks successfully.


Maxusmusti (approved):
LGTM

Issue #33, opened 2022-11-21
SDK Info Exposure Rework (status/details, wait, error-handling, etc.)


Maxusmusti:
Merge both `cluster.status()` and `cluster.is_ready()` into one one function (likely still called `status`). This will tell the user exactly where in the process of setup their cluster is currently (whether still in AppWrapper stages or in Ray stages). Then, there will be a second function called `cluster.details()` that will output the cluster information like all of the specs, worker count, uri, active/inactive, etc. (what we currently see when calling `cluster.status()` on a fully set-up cluster).

Maxusmusti:
Also add a `wait()` function (likely a simple loop checking using the above mentioned `status()` function)

Maxusmusti:
Change `status()` return from bool to info object

thinkahead:
Please ignore my previous comment, it works with the 0.5.0 (I was using 0.4.0 previously)

PR #63, opened 2023-02-13
SDK Info Rework (status, details, wait, etc.)

Maxusmusti:
SDK Info Rework:
 - New all-encpompassing `cluster.status()` method
 - New `cluster.details()` method
 - New `cluster.wait_ready()` method
 - New `CodeFlareStatus.STARTING` status
 - Functions`get_current_namespace()`, `list_all_clusters()`, and `list_all_queued()` now all function as intended
 - A ton of bugfixes and error-handling for both regular use and edge-cases (not logged in, login expired, no resources created yet, resource failure, cluster setup in-progress, etc.)
 - A LOT more testing (up to 95%)
 - Improved messaging on both expected unexpected results

Tied to issue #33

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -103,7 +103,7 @@ class Cluster:
         try:
             with oc.project(namespace):
                 oc.invoke("apply", ["-f", self.app_wrapper_yaml])
-        except oc.OpenShiftPythonException as osp:
+        except oc.OpenShiftPythonException as osp:  # pragma: no cover
             error_msg = osp.result.err()
             if "Unauthorized" in error_msg:
                 raise PermissionError(
@@ -120,7 +120,7 @@ class Cluster:
         try:
             with oc.project(namespace):
                 oc.invoke("delete", ["AppWrapper", self.app_wrapper_name])
-        except oc.OpenShiftPythonException as osp:
+        except oc.OpenShiftPythonException as osp:  # pragma: no cover
             error_msg = osp.result.err()
             if (
                 'the server doesn\'t have a resource type "AppWrapper"' in error_msg
@@ -279,9 +279,12 @@ def get_current_namespace() -> str:
     """
     try:
         namespace = oc.invoke("project", ["-q"]).actions()[0].out.strip()
-    except oc.OpenShiftPythonException as osp:
+    except oc.OpenShiftPythonException as osp:  # pragma: no cover
         error_msg = osp.result.err()
-        if "do not have rights" in error_msg:
+        if (
+            "do not have rights" in error_msg
+            or "Missing or incomplete configuration" in error_msg
+        ):
             raise PermissionError(
                 "Action not permitted, have you run auth.login() or cluster.up()?"
             )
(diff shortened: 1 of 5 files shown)


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -200,45 +189,111 @@ def is_ready(self, print_to_console: bool = True):  # pragma: no cover
...
+            )
+        else:
+            raise osp
     return namespace
 
 
-def list_all_clusters(
-    namespace: str, print_to_console: bool = True
-):  # pragma: no cover
+def list_all_clusters(namespace: str, print_to_console: bool = True):
```
MichaelClifford, line 296: Should this be `list_all_ray_clusters`?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -302,23 +393,42 @@ def _get_ray_clusters(namespace="default") -> List[RayCluster]:  # pragma: no co
...
+            )
+        else:
+            raise osp
 
     for item in app_wrappers:
         app_wrapper = _map_to_app_wrapper(item)
         if filter and app_wrapper.status in filter:
             list_of_app_wrappers.append(app_wrapper)
         else:
+            # Unsure what the purpose of the filter is
```
MichaelClifford, line 421: I think that this would allow you to specify which types of statuses you want to get. 

But I also don't see where this function is used. It doesn't seem to get called elsewhere in this file.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -248,7 +303,7 @@ def list_all_clusters(
     return clusters
 
 
-def list_all_queued(namespace: str, print_to_console: bool = True):  # pragma: no cover
+def list_all_queued(namespace: str, print_to_console: bool = True):
```
MichaelClifford, line 306: Should this function ignore running clusters? And only show Pending clusters?

MichaelClifford:
@Maxusmusti Looks good :+1:  Just had a couple comments about function names before we merge. Let me know what you think.

MichaelClifford:
@Maxusmusti  also, do we want to update the notebooks showing how to use these new features in this PR? Or included it in a later PR?

Maxusmusti:
> @Maxusmusti also, do we want to update the notebooks showing how to use these new features in this PR? Or included it in a later PR?

I can include those in a separate PR that updates the documentation and prepares for new release

src/codeflare_sdk/cluster/cluster.py
Maxusmusti: Yeah, my confusion is why it does the same exact thing in the if and the else, makes the if statement, and filter, useless (unless I'm just missing something)

src/codeflare_sdk/cluster/cluster.py
Maxusmusti: I think it is fine for now, just so that it shows all active AppWrappers/MCAD-managed resources. I think a good change for the future would have the use be able to pass in the "filter" that this function uses internally in a clearer manner, and then actually fix it functionally within by ignoring what doesn't match the filter.

src/codeflare_sdk/cluster/cluster.py
Maxusmusti: I think it's left generic for "future-proofing", as keeping in line with why we don't mention ray in other user-facing functions, we could theoretically have other types of backends in the future, so it's more so meant to list all "codeflare clusters". Do you think this would lead to any unecessary confusion though? If so, we could change it to `list_all_ray_clusters` for now, or even `list_all_codeflare_clusters`?

src/codeflare_sdk/cluster/cluster.py
Maxusmusti: That way a user could see all pending requests, all failed requests, all running requests, or any combination.

src/codeflare_sdk/cluster/cluster.py
MichaelClifford: Ok, that makes sense. I think its fine for now then to keep it general.  I think this is just an artifact of the term "cluster" being a bit overloaded here. Reading through the code, we are basically tracking two types of things, AppWrappers and Ray Clusters. And the term "cluster" seemed somewhat vague. But, something we can maybe address more systematically in the future.

MichaelClifford (approved):
LGTM

Issue #32, opened 2022-11-21
Update SDK to manage jobs


MichaelClifford:
As a user of the the codeflare-sdk I want to be able to mange the submission of jobs from the sdk, including torchx jobs. 

* Create a generic Jobs class that defines a job to be submitted to the cluster. 
*  Add a specific jobs class to manage torchx jobs.

PR #64, opened 2023-02-14
New 32

KPostOffice:
Please feel free to comment and offer suggestions.

Closes: #32


src/codeflare_sdk/jobs/jobs.py
```diff
@@ -0,0 +1,135 @@
...
+                cpu=cluster.config.max_cpus,
+                gpu = cluster.config.gpu,
+                memMB = 1024 * cluster.config.max_memory,  # cluster memory is in GB
+                j=j,
+                env=job_definition.config.env,
+                debug=job_definition.config.debug  # TODO: expose
+                # max_retries=0,  # default
+                # mounts=None,  # default
+            ),
+            scheduler="ray",  # can be determined by type of cluster if more are introduced
```
MichaelClifford, line 109: Should we add this as configurable parameter, "scheduler" in the Config, with 'ray' as the default. And then rename this class `TorchXJob()`? I know the status() and logs() functions assume ray at the moment, but the class name and hard coding  feel a bit awkward here. WDYT?

KPostOffice: Yeah, the scheduler is really dependent on the cluster, so I don't think it needs to be exposed to the user. It shouldn't be hard coded though.

src/codeflare_sdk/jobs/jobs.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        self.job_definition: TorchXJobDefinition = job_definition
+        self.cluster: "Cluster" = cluster
+        self._app_handle = torchx_runner.schedule(
+            job_definition._dry_run(cluster, *script_args)
+        )
+        all_jobs.append(self)
+
+    @property
+    def job_id(self):
```
MichaelClifford, line 145:
```suggestion
    def job_id(self):
    """
    Returns the job id if present. If not returns job id from ray dashboard. 
    """
```

src/codeflare_sdk/jobs/jobs.py
```diff
@@ -0,0 +1,162 @@
...
+        if hasattr(self, "_job_id"):
+            return self._job_id
+        dashboard_address = f"{self.cluster.cluster_dashboard_uri(self.cluster.config.namespace).lstrip('http://')}:8265"
+        _, _, job_id = parse_app_handle(self._app_handle)
+        self._job_id = job_id.lstrip(f"{dashboard_address}-")
```
MichaelClifford, line 150:
```suggestion
        if hasattr(self, "_job_id"):
            return self._job_id
        dashboard_address = f"{self.cluster.cluster_dashboard_uri(self.cluster.config.namespace).lstrip('http://')}:8265"
        _, _, job_id = parse_app_handle(self._app_handle)
        self._job_id = job_id.lstrip(f"{dashboard_address}-")
        return self._job_id
```

src/codeflare_sdk/jobs/jobs.py
```diff
@@ -0,0 +1,162 @@
...
+The jobs sub-module contains the definitions for the Job objects, which represent
+the methods by which a user can submit a job to a cluster, check the jobs status and 
+access the jobs logs. 
+"""
+
+import abc
+from typing import List
+from pathlib import Path
+
+from ray.job_submission import JobSubmissionClient
```
MichaelClifford, line 25: I don't think this is needed.

src/codeflare_sdk/jobs/jobs.py
```diff
@@ -0,0 +1,162 @@
...
+        dashboard_address = f"{cluster.cluster_dashboard_uri(cluster.config.namespace).lstrip('http://')}"
+        return torchx_runner.dryrun(
+            app=ddp(
+                *script_args,
+                script=self.config.script,
+                m=self.config.m,
+                name=self.config.name,
+                h=None,  # for custom resource types
+                cpu=cluster.config.max_cpus,
+                gpu=cluster.config.gpu,
```
MichaelClifford, line 102: Can we have these be the defaults, but expose them as options to the user. This setup assume you want to use 100% of cluster resources for every job submitted. Not sure if this would always be the case.

MichaelClifford (changes requested):

KPostOffice:
I'm having trouble balancing making the interface simple and making the code sufficiently generic. As I make it more generic I'm just falling back to the torchx design, and I feel like the abstraction isn't offering much in terms of simplification. I think I might need a rethink.

KPostOffice:
Superseded by #70

PR #66, opened 2023-02-21
[WIP - Do Not Merge] Separate auth methods from cluster methods

MichaelClifford:
This PR is aimed at making working with OpenShift environments a bit more user friendly by separating the authentication process from the the cluster object. Previously we automatically logged users in and out along with the cluster.up and cluster.down commands. This introduced the problem of needing to get a new token each time they restarted their ray cluster, since it also logged them out of their OpenShift cluster. Now users can use auth.login and auth.logout to control cluster access. cluster.up and cluster.down no longer force the authentication action, and the authentication object is no longer required in the cluster config.

### DIFF
--- a/demo-notebooks/batch-job/batch_mnist.ipynb
+++ b/demo-notebooks/batch-job/batch_mnist.ipynb
@@ -24,7 +24,9 @@
     "    token = \"XXXX\",\n",
     "    server = \"XXXX\",\n",
     "    skip_tls=True\n",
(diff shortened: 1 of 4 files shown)


MichaelClifford:
cc @jbusche

KPostOffice (commented):
/lgtm

MichaelClifford:
This PR needs to be reworked to account for changes made in #63

PR #69, opened 2023-03-02
Separate auth from cluster methods

MichaelClifford:
This PR is aimed at making working with OpenShift environments a bit more user friendly by separating the authentication process from the the cluster object. Previously we automatically logged users in and out along with the cluster.up and cluster.down commands. This introduced the problem of needing to get a new token each time they restarted their ray cluster, since it also logged them out of their OpenShift cluster. Now users can use auth.login and auth.logout to control cluster access. cluster.up and cluster.down no longer force the authentication action, and the authentication object is no longer required in the cluster config.

(This is an update of #66)

### DIFF
--- a/src/codeflare_sdk/cluster/auth.py
+++ b/src/codeflare_sdk/cluster/auth.py
@@ -74,6 +74,8 @@ class TokenAuthentication(Authentication):
             error_msg = osp.result.err()
             if "The server uses a certificate signed by unknown authority" in error_msg:
                 return "Error: certificate auth failure, please set `skip_tls=True` in TokenAuthentication"
+            elif "invalid" in error_msg:
+                raise PermissionError(error_msg)
             else:
                 return error_msg
         return response.out()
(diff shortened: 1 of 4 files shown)


src/codeflare_sdk/cluster/auth.py
```diff
@@ -82,7 +84,8 @@ def logout(self) -> str:
         """
         This function is used to logout of an OpenShift cluster.
         """
-        response = oc.invoke("logout")
+        args = [f"--token={self.token}", f"--server={self.server}:6443"]
```
Maxusmusti, line 87: LGTM, just curious why the logout function needs the server/token?

MichaelClifford: @Maxusmusti, while working on this I ran into a few edge cases where the logout function failed because the connection to the cluster had been interrupted. The addition of the token and server just makes the logout function more stable. 

If you don't like this we could replace it with some different error handling statements? Let me know what you think.

Maxusmusti (approved):
LGTM

PR #74, opened 2023-03-13
Remove hard-coded port 6443 in auth

MichaelClifford:
Based on user feedback, we realized that hard coding the OpenShift server port was causing issues with some users preventing them from being able to log in. 

This PR simple removes the `:6443` suffix from server parameter so that the port must be explicitly included. 

We can add future PR's to include some more sophisticated server address parsing and error handling.


anishasthana (commented):
/lgtm

anishasthana (approved):
We should follow up with a PR later to make the server string logic a little more intelligent, something like 
```
if len(self.server.split(":")) > 1:
    # Use server string
else:
   # User server string + :6443
```

anishasthana (approved):
/lgtm

PR #78, opened 2023-03-27
Adding TorchX-MCAD Scheduler Support to Jobs

Maxusmusti:
IN PROGRESS: Adding the `kubernetes_mcad` scheduler option to TorchX jobs via the SDK

### DIFF
--- a/src/codeflare_sdk/job/jobs.py
+++ b/src/codeflare_sdk/job/jobs.py
@@ -61,6 +61,7 @@ class DDPJobDefinition(JobDefinition):
         mounts: Optional[List[str]] = None,
         rdzv_port: int = 29500,
         scheduler_args: Optional[Dict[str, str]] = None,
+        image: Optional[str] = None,
     ):
         if bool(script) == bool(m):  # logical XOR
             raise ValueError(
@@ -82,6 +83,7 @@ class DDPJobDefinition(JobDefinition):
         self.scheduler_args: Dict[str, str] = (
             scheduler_args if scheduler_args is not None else dict()
         )
+        self.image = image
 
     def _dry_run(self, cluster: "Cluster"):
         j = f"{cluster.config.max_worker}x{max(cluster.config.gpu, 1)}"  # # of proc. = # of gpus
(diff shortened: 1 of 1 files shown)


src/codeflare_sdk/job/jobs.py
```diff
@@ -108,15 +110,58 @@ def _dry_run(self, cluster: "Cluster"):
...
+    def _missing_spec(spec: str):
+        raise ValueError(f"Job definition missing arg: {spec}")
+
+    def _dry_run_no_cluster(self):
+        return torchx_runner.dryrun(
+            app=ddp(
+                *self.script_args,
+                script=self.script,
+                m=self.m,
+                name=self.name,
```
MichaelClifford, line 122: Name is also a required argument here. please add _missing_spec()

MichaelClifford (changes requested):

src/codeflare_sdk/job/jobs.py
Maxusmusti: Ah, is this only the case with `kubenetes_mcad` (is it fine to remain optional with `ray`)?

Maxusmusti: Also odd that it is required when submitting via python, but they don't make you use `--name` via CLI

src/codeflare_sdk/job/jobs.py
```diff
@@ -108,15 +110,58 @@ def _dry_run(self, cluster: "Cluster"):
...
+    def _missing_spec(self, spec: str):
+        raise ValueError(f"Job definition missing arg: {spec}")
+
+    def _dry_run_no_cluster(self):
+        return torchx_runner.dryrun(
+            app=ddp(
+                *self.script_args,
+                script=self.script,
+                m=self.m,
+                name=self.name if self.name is not None else self._missing_spec("name"),
```
MichaelClifford, line 122: total nit, do you wan this on one line like this instead of
```
name=self.name
if self.name is not None
else self._missing_spec("name")

```

src/codeflare_sdk/job/jobs.py
MichaelClifford: Let me double check, but I'm pretty sure Ray gets the "name" another way to its not a required arg here.

MichaelClifford: > Also odd that it is required when submitting via python, but they don't make you use --name via CLI
Can look into this too.

src/codeflare_sdk/job/jobs.py
Maxusmusti: Ah, I just run `black` to auto-format, guessing that line wasn't deemed "too long", but I can make that uniform with the rest

MichaelClifford (approved):

PR #79, opened 2023-03-30
Add Job Cancelling and More Unit Tests

Maxusmusti:
Added:
 - Job cancelling + unit test
 - More unit tests for torchx-mcad jobs

### DIFF
--- a/src/codeflare_sdk/job/jobs.py
+++ b/src/codeflare_sdk/job/jobs.py
@@ -169,3 +169,6 @@ class DDPJob(Job):
 
     def logs(self) -> str:
         return "".join(torchx_runner.log_lines(self._app_handle, None))
+
+    def cancel(self):
+        torchx_runner.cancel(self._app_handle)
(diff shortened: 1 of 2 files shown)


src/codeflare_sdk/job/jobs.py
```diff
@@ -169,3 +169,6 @@ def status(self) -> str:
 
     def logs(self) -> str:
```
KPostOffice, line 170: Did you test the logging functionality with a notebook. I think it might break with MCAD logs due to the `None` as the second arg.

KPostOffice (commented):
Everything else looks good at first glance. Just had a question.

KPostOffice:
Can you also make cluster default to None so the user isn't required to pass any args?

https://github.com/project-codeflare/codeflare-sdk/blob/826630703841fdda12ff624b565faffbe0ac8e33/src/codeflare_sdk/job/jobs.py#L156

MichaelClifford (approved):

PR #85, opened 2023-04-06
Change default namespace logic

anishasthana:
- Update developer instructions
- Change Default namespace logic to use user's current namespace

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ dist/
 .python-version
 __pycache__/
 .coverage
+Pipfile
+Pipfile.lock
--- a/README.md
+++ b/README.md
@@ -15,26 +15,31 @@ Can be installed via `pip`: `pip install codeflare-sdk`
 
 ## Development
 
-For testing, make sure to have installed:
- - `pytest`, `pytest-mock` (can both be installed with `pip`)
- - The remaining dependencies located in `requirements.txt`
- - To run the unit tests, run `pytest -v tests/unit_test.py`)
- - Any new test functions/scripts can be added into the `tests` folder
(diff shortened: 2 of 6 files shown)


MichaelClifford:
Ok, kind of annoying, but I think we need to change this PR a bit to account for the fact that using the MCAD scheduler also needs a `get_namespace` call to make sure it doesn't just use the "default" namespace. 

https://github.com/project-codeflare/codeflare-sdk/blob/f529763df746ab26a2c8519f7df15c5a3b2de44a/src/codeflare_sdk/job/jobs.py#L146-L149

We need to add something here like:

`self.scheduler_args["namespace"] = openshift.get_project()`

We also need to deal with the fact that we already have a `get_current_namespace()` function within the `cluster.py` file. (This is an artifact from when we assumed a cluster object would always be required).  
 
My suggestion is that we delete `get_current_namespace()` as its not used by any other function, and then rely on the openshift package to call `openshift.get_project()` wherever we need the project name (basically your idea from yesterday :smile:).  This avoids the need for a higher level `utils.py` type file to house a shared function like `get_current_namespace`. 

Let me know what you think.

anishasthana:
>My suggestion is that we delete get_current_namespace() as its not used by any other function, and then rely on the openshift package to call openshift.get_project() wherever we need the project name (basically your idea from yesterday 😄). This avoids the need for a higher level utils.py type file to house a shared function like get_current_namespace.

Question about that -- yesterday we also said that there is a need to allow users to determine what namespace they are in from the API. How would we structure things for that?

MichaelClifford (approved):
thanks @anishasthana 
LGTM!

PR #87, opened 2023-04-07
add rdzv_backend parameter to DDPJobDefinition

MichaelClifford:
This change allows the user to define the rdzv_backend to be used when creating a DDPJobDefinition. 

This change is required because currently, the backend is hard coded to "static" in [torchx dist.py](https://github.com/project-codeflare/torchx/blob/7eac182893d237c22ef102fadea093da6f0626f8/torchx/components/dist.py#L149). Howver, depending on whether or not we are using the Ray Scheduler or the MCAD scheduler, the required backend needs to change as well. 

This PR will be paired with an upcoming change to torchx/OCP that accounts for the different backends submitted here.


anishasthana:
@MichaelClifford Can you rebase on main?

MichaelClifford:
@anishasthana rebased :)

Maxusmusti (approved):
LGTM

PR #88, opened 2023-04-10
Update python tests job

anishasthana:


.github/build/requirements.txt
```diff
@@ -0,0 +1,11 @@
+openshift-client==1.0.18
+rich==12.5.1
+ray[default]==2.1.0
+kubernetes==26.1.0
+codeflare-torchx==0.5.0.dev5
```
KPostOffice, line 5: I think it's this. Old version of the custom torchx. It should be 0.6.0.dev0

anishasthana: Oh good catch, thanks!

.github/build/Containerfile
```diff
@@ -5,10 +5,12 @@ LABEL summary="Toolchain for running pre-commit hooks." \
     io.k8s.display-name="Pre-Commit Toolchain"
 
 USER root
-RUN dnf install nodejs
+RUN dnf install nodejs -y && \
+    dnf clean all && \
+    rm -rf /var/cache/dnf
```
KPostOffice, line 10: Need to install the openshift client cli as it is used directly by the  `openshift-client` python library

.github/build/requirements.txt
KPostOffice, line 1: Could we just change the build context when running the `Dockerfile` so that we don't have to maintain two requirements files?

anishasthana: The build was failing -- it was unable to access files outside the same directory.

MichaelClifford (approved):
LGTM

requirements.txt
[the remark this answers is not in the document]
KPostOffice: The idea is that one of them has requirements needed just to run it and the other has developer requirements. For example, there's no reason to install `pytest` or `black` in the final image.

.github/build/requirements.txt
KPostOffice: We could put the Container file in the root of the repo couldn't we?

.github/build/Containerfile
```diff
@@ -5,10 +5,16 @@ LABEL summary="Toolchain for running pre-commit hooks." \
...
+ADD https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz $TMPDIR/
```
jiripetrlik, line 11: I think that if we use ADD here, the Docker will create a separate image layer with tar file. And even if we delete it in the next command it will stay in the image. It will be only hidden by next layer. I would suggest to download tar with oc as part of the RUN command on next line.

jiripetrlik (commented):
I have one minor comment.

anishasthana:
Replaced by #305

PR #96, opened 2023-04-25
Add use of priorityClasses to SDK

### LINKED ISSUE #182: Support use of PriorityClass for Appwrappers in the SDK
One feature of MCAD is that is allows for priority and preemption of jobs. The SDK needs to support this feature by allowing a user to define a priority that updates both the priorityClassName and the spec priority that gets added to the generated appwrapper (see this PR for an example AppWrapper #96).

However, the priorityClassNames are non-standard and will be defined and occasionally edited by the cluster admin, therefore the SDK needs to lookup the available priorityClassNames available on the cluster and expose them to the user so that they can use the correct priority class, and raise an error when they try to use on that's not available. 

The SDK also need to handle the case where there are no relevant priority classes on the cluster.

MichaelClifford:
This PR enables the use of `PriorityClasses` for our RayClusters. 

closes #182 

* Adds a new parameter `dispatch_priority` to `ClusterConfig`
* If `dispatch_priority` is left empty, appwrapper is not affected. 
* If `dispatch_priority` is used, then appwrapper adds the field `priorityClassName` to the head and worker group specs. 
* Also adds functionality to check the validity of the appwrapper during `cluster.up()`. This addition is required as not all values for `dispatch_priority` are valid and the cluster needs to be checked for valid values before applying the appwrapper.

### DIFF
--- a/src/codeflare_sdk/templates/base-template.yaml
+++ b/src/codeflare_sdk/templates/base-template.yaml
@@ -114,7 +114,6 @@ spec:
                           operator: In
                           values:
                           - "aw-kuberay"
-                priorityClassName: "default-priority"
                 containers:
                 # The Ray head pod
                 - env:
@@ -224,7 +223,6 @@ spec:
                           operator: In
                           values:
                           - "aw-kuberay"
-                priorityClassName: "default-priority"
                 initContainers:
                 # the env var $RAY_IP is set by the operator if missing, with the value of the head service name
                 - name: init-myservice
--- a/src/codeflare_sdk/templates/base-template.yaml
+++ b/src/codeflare_sdk/templates/base-template.yaml
@@ -8,6 +8,8 @@ metadata:
     orderedinstance: "m4.xlarge_g4dn.xlarge"
 spec:
   priority: 9
+  schedulingSpec:
+    minAvailable: 3
   resources:
     Items: []
     GenericItems:
@@ -112,6 +114,7 @@ spec:
                           operator: In
                           values:
                           - "aw-kuberay"
+                priorityClassName: "default-priority"
                 containers:
                 # The Ray head pod
                 - env:
@@ -221,6 +224,7 @@ spec:
                           operator: In
                           values:
                           - "aw-kuberay"
+                priorityClassName: "default-priority"
                 initContainers:
                 # the env var $RAY_IP is set by the operator if missing, with the value of the head service name
                 - name: init-myservice
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -372,7 +372,7 @@ def generate_appwrapper(
     env,
     local_interactive: bool,
     image_pull_secrets: list,
-    priority: str,
+    dispatch_priority: str,
 ):
     user_yaml = read_template(template)
     appwrapper_name, cluster_name = gen_names(name)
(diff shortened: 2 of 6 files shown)


src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -314,6 +343,12 @@ def main():  # pragma: no cover
         default="default",
         help="Set the kubernetes namespace you want to deploy your cluster to. Default. If left blank, uses the 'default' namespace",
     )
+    parser.add_argument(
+        "--priority",
+        required=False,
+        default="low",
+        help="Set the priority of the cluster. Default is 'low'. Options are 'low', 'default', 'high'",
```
asm582, line 350: Nit change cluster to framework cluster

asm582, line 350: default should be default priority

asm582, line 350: We should mention the word `dispatch priority`

asm582: and to be consistent in the code change variable name priority to dispatch priority and also the args name

asm582 (changes requested):

MichaelClifford:
@asm582 Do you want to give this another look and approve the changes requested?

src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -76,6 +76,27 @@ def update_labels(yaml, instascale, instance_types):
...
 
+def update_priority(yaml, item, priority):
+    if priority not in ["low", "default", "high"]:
+        sys.exit("Priority must be 'low', 'default', or 'high'")
+
+    priority_levels = {
+        "low": (1, "low-priority"),
+        "default": (5, "default-priority"),
+        "high": (10, "high-priority"),
+    }
```
KPostOffice, line 87: I think this should be a global constant.

src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -76,6 +76,27 @@ def update_labels(yaml, instascale, instance_types):
         metadata.pop("labels")
 
 
+def update_priority(yaml, item, priority):
+    if priority not in ["low", "default", "high"]:
```
KPostOffice, line 80: This can be changed to:
```python
if priority not in PRIORITY_LEVELS:
    ...
```

MichaelClifford:
Thanks for the comments @KPostOffice . updated :smile: 

@asm582 Do you want to give this another look and resolve any outstanding comments :smile:

asm582:
LGTM, priority levels for now are hardcoded in some way, in the future users may need fewer or more priority levels.

asm582 (approved):

src/codeflare_sdk/templates/base-template.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
     orderedinstance: "m4.xlarge_g4dn.xlarge"
 spec:
   priority: 9
+  schedulingSpec:
+    minAvailable: 3
```
Maxusmusti, line 12: Just curious, what is the `minAvailable` field for/mean?

tedhtchang: [Expected number of pods in running and/or completed state. Requeuing is triggered
when the number of running/completed pods is not equal to this value. When not specified,
requeuing is disabled and no check is performed.](https://github.com/project-codeflare/multi-cluster-app-dispatcher/blob/9035cd73d5bd03a1b584b3567255f6af5006e7d6/config/crd/bases/mcad.ibm.com_appwrappers.yaml#LL262C27-L262C27)

asm582:
@MichaelClifford Checked this is approved, please feel free to merge

Maxusmusti (approved):
LGTM

PR #144, opened 2023-06-05
fix error preventing NB users from installing from pip

### LINKED ISSUE #126: Update CodeFlare notebook image to allow users to pip install additional packages.
The current codeflare notebook images appears to limit users ability to add additional python packages to their environment via `pip install ...` 

image

This error happens for some, but not all packages. 

This image is currently based on `quay.io/opendatahub/notebooks:jupyter-minimal-ubi8-python-3.8-4c8f26e` 

We need to develop a solution so that users can update their notebook images with pip.

MichaelClifford:
Resolves #126

In this PR we update the permissions during the image build so that users no longer get an error when trying to install some packages via pip.


anishasthana (approved):
Can confirm that the image works. :ship:

PR #151, opened 2023-06-08
Update CodeFlare-sdk Notebook Image to 0.4.4

anishasthana:


custom-nb-image/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN pip install -r requirements.txt
 RUN pip uninstall pickle5 -y
 
 # Install codeflare-sdk and other libraries
-RUN pip install codeflare-sdk==0.4.3 \
```
KPostOffice, line 26: https://docs.docker.com/build/guide/build-args/

Can we use build args here? If we do this then we can use the build arg as part of our release process after the library has been published that way we don't have to keep updating the version in this file whenever we create a new release.

Maxusmusti (approved):
LGTM

custom-nb-image/Dockerfile
anishasthana: I don't disagree. I've opened #152  to address that later.

PR #157, opened 2023-06-13
Create GitHub action to automate release

### LINKED ISSUE #148: Create GitHub action to automate codeflare-sdk release
The action should perform all steps described in https://github.com/project-codeflare/codeflare-sdk/pull/141.

jiripetrlik:
Fixes #148


.github/workflows/release.yaml
```diff
@@ -0,0 +1,64 @@
+name: Release
+
+on:
+    workflow_dispatch:
+        inputs:
```
sutaakar, line 5: Can you set the inputs as `required`?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,64 @@
...
+              with:
+                commit_message: "New release: ${{ github.event.inputs.new_version }}"
+            - name: Create Github release
+              uses: ncipollo/release-action@v1
+              with:
+                  tag: "v${{ github.event.inputs.new_version }}"
+            - name: Update SDK version in the Docker image
+              run: sed -i "s/\(.*codeflare-sdk==\)[^\\s]*\(.*\)/\${{ github.event.inputs.new_version }}\2/" custom-nb-image/Dockerfile
+            - name: Update Docker image version
+              run: echo ${{ github.event.inputs.new_version }} > custom-nb-image/VERSION
```
sutaakar, line 60: Based on the readme the image should be built and pushed?

.github/workflows/release.yaml
jiripetrlik: Added required: true property for all parameters.

.github/workflows/release.yaml
jiripetrlik: It will automatically trigger the "image" workflow by commit with changes of the VERSION file.  The "image" workflow will take care about the image build.

anishasthana: Can we update that notebook image github action to be triggered on repo tag instead of the version file? @Maxusmusti wdyt?

sutaakar: The approach I took for InstaScale is to trigger the action manually, tag is created as part of github action execution.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,76 @@
...
+              with:
+                  tag: "v${{ github.event.inputs.new_version }}"
+
+            - name: Update SDK version in the Docker image
+              run: sed -i "s/\(.*codeflare-sdk==\)[^\\s]*\(.*\)/\1${{ github.event.inputs.new_version }}\2/" custom-nb-image/Dockerfile
+            - name: Image Build
+              run: |
+                cd custom-nb-image
+                docker build -t quay.io/project-codeflare/notebook:${{ github.event.inputs.image_tag }} .
+                docker tag quay.io/project-codeflare/notebook:${{ github.event.inputs.image_tag }} quay.io/project-codeflare/notebook:latest
```
sutaakar, line 66: Can you separate creation of latest tag to be optional, like in https://github.com/project-codeflare/instascale/blob/e3df822968e6617cf5e2bcc60469268a42f55738/.github/workflows/instascale-release.yml#L62-L66 ?

The reason is that some releases may not be latest one (i.e. some patch release).

.github/workflows/release.yaml
```diff
@@ -0,0 +1,76 @@
+name: Release
+
+on:
+    workflow_dispatch:
+        inputs:
+            new_version:
```
sutaakar, line 6: Can you provide a sample values for parameters like in https://github.com/project-codeflare/instascale/blob/e3df822968e6617cf5e2bcc60469268a42f55738/.github/workflows/instascale-release.yml#L11 ?

sutaakar:
@jiripetrlik is it possible to test whole workflow on own fork? If so, what all needs to be done/configured?

.github/workflows/release.yaml
jiripetrlik: Done. :+1:

README.md
```diff
@@ -62,13 +62,10 @@ The following instructions apply when doing release manually.
 * If not present already, add the API token to Poetry.
 `poetry config pypi-token.pypi API_TOKEN`
 * Publish the Python package. `poetry publish`
```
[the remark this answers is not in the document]
jiripetrlik: The instructions are for the situations when you are doing release manually, but I put there a note that there is a workflow which does it automatically.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,76 @@
...
+              type: string
+              required: true
+            image_tag:
```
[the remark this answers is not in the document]
jiripetrlik: Unified to use "release-version". It is a little bit complicated as Github and Docker images have "v" prefix, but Pypi does not.

.github/workflows/release.yaml
jiripetrlik: Done. :+1:

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
...
+            - name: Install Poetry
+              uses: abatilo/actions-poetry@v2
+              with:
+                poetry-version: ${{ github.event.inputs.poetry_version }}
+            - name: Run poetry install
+              run: poetry install --with docs
+            - name: Run poetry build
+              run: poetry build
+            - name: Create new documentation
+              run: poetry run pdoc --html -o docs src/codeflare_sdk && pushd docs && rm -rf cluster job utils && mv codeflare_sdk/* . && rm -rf codeflare_sdk && popd
```
Maxusmusti, line 51: Is there a reason this does a `poetry run pdoc` instead of just directly running `pdoc`?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
+name: Release
+
+on:
+    workflow_dispatch:
+        inputs:
+            release-version:
+              type: string
+              required: true
+              description: 'Version number'
+              default: '0.1.0'
```
Maxusmusti, line 10: Should we have a default version at all? It seems like it should never be used (we never want to release `0.1.0` again), so if it is ever used something has gone wrong

.github/workflows/release.yaml
jiripetrlik: It will run pdoc in the virtual environment which Poetry created and there are all installed dependencies. I think if you run just "pdoc" it will reply with command not found.

MichaelClifford (changes requested):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,84 @@
...
+              description: 'Select if the built image should be tagged as latest'
+              required: true
+              type: boolean
+            quay-organization:
+              description: 'Quay organization used to push the built images to'
+              required: true
+              default: 'project-codeflare'
+            python_version:
+              type: string
+              default: "3.10"
```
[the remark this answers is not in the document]
jiripetrlik: @MichaelClifford 
Good point. I've changed the version of Python to 3.8.

MichaelClifford (approved):
LGTM!
Thanks for the changes

sutaakar (approved):

Maxusmusti (approved):
LGTM

KPostOffice (approved):

Issue #161, opened 2023-06-14
Add support for specifying a cluster CA certificate to the sdk


accorvin:
We currently [support](https://github.com/project-codeflare/codeflare-sdk/blob/3b41a22dec1b2471bdc46d8679dae6a85eb96b39/src/codeflare_sdk/cluster/auth.py#LL60C33-L60C33) the `skip_tls` argument in the SDK authentication object for clusters with untrusted CA certificates. This is fine for development purposes, but will be unacceptable for production workloads for most customers.

We should update the SDK to allow specifying the path to a CA certificate bundle so that the auth endpoint can be trusted.

I think we could additionally auto configure trust of the cluster based on where we can expect the cluster CA cert to be stored on disk when running in an openshift pod.

Bobbins228:
Closing this as I have opened another issue relating to authentication via Kubernetes instead of OpenShift. #185

Bobbins228:
Closing this as I have opened another issue relating to authentication via Kubernetes instead of OpenShift. https://github.com/project-codeflare/codeflare-sdk/issues/185

PR #164, opened 2023-06-19
Add support for specifying a cluster CA certificate to the sdk

Bobbins228:
Linked to [issue #161](https://github.com/project-codeflare/codeflare-sdk/issues/161)
By setting `skip_tls=False` the login command will use the arg `--certificate-authority` with the specified path to a cert a user sets with `ca_cert_path`

Example:
```
# Create authentication object for oc user permissions
auth = TokenAuthentication(
    token = "XXXXX",
    server = "XXXXX",
    ca_cert_path = "<path-to-cert.crt>"
)
auth.login()

### DIFF
--- a/src/codeflare_sdk/cluster/auth.py
+++ b/src/codeflare_sdk/cluster/auth.py
@@ -49,7 +49,13 @@ class TokenAuthentication(Authentication):
     cluster when the user has an API token and the API server address.
     """
 
-    def __init__(self, token: str = None, server: str = None, skip_tls: bool = False):
+    def __init__(
+        self,
+        token: str = None,
+        server: str = None,
+        ca_cert_path: str = None,
+        skip_tls: bool = False,
+    ):
         """
         Initialize a TokenAuthentication object that requires a value for `token`, the API Token
         and `server`, the API server address for authenticating to an OpenShift cluster.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by tedhtchang — COMMENTED** [review]
@Bobbins228 Thanks. Looks good. Could you provide an example how to test/verify this PR ?

**Comment by Bobbins228** [comment]
> @Bobbins228 Thanks. Looks good. Could you provide an example how to test/verify this PR ?

Hi Ted,

To begin clone this repo and checkout the `add-ca-support` branch.
Use `poetry build` to create a new whl file located in the `dist` folder.
On a Jupytr notebook drag the `whl` file to your working directory and install it with `!pip install <whl-file>.whl`.

import the library with 
```
from codeflare_sdk.cluster.cluster import Cluster, ClusterConfiguration
from codeflare_sdk.cluster.auth import TokenAuthentication
```

In another cell log in using...
```
auth = TokenAuthentication(
    token = "XXXXX",
    server = "XXXXX",
    skip_tls=False,
    ca_cert_path="<path>"
)
auth.login()
```

While testing this out I found that the the Jupytr Notebook pod has a trusted cert bundle located at `/etc/pki/tls/certs/ca-bundle.crt`.
I could set this as the default path value instead of the `None` I have already.

Another thing I noticed while testing is that with my work in place you can no longer log in with just `token` and `server` unless `skip_tls=True`. Is this okay? 

When you try to log in without specifying a path with `skip_tls=False` you get this  `'error: open None: no such file or directory\n'`. We could set the default path to the default cert path mentioned above to mitigate this.

Bobbins228:
Closing this as we have moved towards using the Kubernetes api. See #186

Issue #152, opened 2023-06-08
Use Build Args in Notebook Image build.


anishasthana:
Refer to https://github.com/project-codeflare/codeflare-sdk/pull/151#discussion_r1223413739 for complete context.

cc @KPostOffice

PR #167, opened 2023-06-20
Added build args for codeflare-sdk in notebook image build

Bobbins228:
Added a build arg for specifying the codeflare-sdk version.
References Issue #152 

To test this run `docker build . --build-arg="SDK_VERSION=<VERSION>"` in the `custom-nb-image` folder.
The default version is set to `0.4.4`


custom-nb-image/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN pip install -r requirements.txt
 RUN pip uninstall pickle5 -y
 
 # Install codeflare-sdk and other libraries
```
KPostOffice, line 25: I think the ARG needs to be moved here so that layers are reused when building.

See: https://stackoverflow.com/questions/41593135/does-putting-arg-at-top-of-dockerfile-prevent-layer-re-use

custom-nb-image/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 FROM  quay.io/opendatahub/notebooks:jupyter-minimal-ubi8-python-3.8-4c8f26e
-
+ARG SDK_VERSION=0.4.4
```
MichaelClifford, line 16: Can we change the default to `0.*` so that we get the most up to date version if none is specified? (We will need to update this once a `1.0` comes out at some point in the future.

MichaelClifford (changes requested):

KPostOffice (dismissed):
Did a quick test and the image seems to be built as expected

KPostOffice (approved):

MichaelClifford (approved):
LGTM!

Thanks for making the changes

PR #179, opened 2023-06-26
add .venv and build directory to git ignore

KPostOffice:
Added poetry build directory to gitignore as well as any python venv directory starting with `.venv`


MichaelClifford (approved):
LGTM!

But needs a rebase now.

Issue #185, opened 2023-06-28
Update authentication for Kubernetes 


Bobbins228:
References #139 

Replace the old OpenShift authentication methods with Kubernetes friendly authentication.
Allow a user to load their own k8s config file.

PR #186, opened 2023-06-28
Updated authentication for Kubernetes

Bobbins228:
# Update authentication for Kubernetes
references #185
#146 
A user can now authenticate using their own k8s config file.

A user can authenticate with their token and server address.

All methods that used `config.load_kube_config()` have been updated to check if a user has already specified their own config file so it isn't overridden.

# How to test the new authentication methods
`git pull https://github.com/Bobbins228/codeflare-sdk.git`
`git checkout update-authentication`
Run `poetry build` within the codeflare-sdk folder to build your own wheel.
Install the sdk with `pip install codeflare_sdk-0.0.0.dev0-py3-none-any.whl`
Import the necessary classes.
```
# Import pieces from codeflare-sdk
from codeflare_sdk.cluster.cluster import Cluster, ClusterConfiguration
from codeflare_sdk.cluster.auth import TokenAuthentication, KubeConfigFileAuthentication
```
## Loading a user's own config file
```
auth = KubeConfigFileAuthentication(
            kube_config_path="/path/to/config"
        )
auth.load_kube_config()
```
## Authenticating with token + server
**Authenticating with certificate**
```
auth = TokenAuthentication(
    token = "TOKEN",
    server = "SERVER",
    skip_tls=False,
    ca_cert_path="path/to/ca_bundle.crt"
)
auth.login()
```
**Authenticating by skipping tls verification**
```
auth = TokenAuthentication(
    token = "TOKEN",
    server = "SERVER",
    skip_tls=True
)
auth.login()
```
**Logging out**
`auth.logout()`


src/codeflare_sdk/cluster/auth.py
```diff
@@ -58,65 +103,176 @@ def __init__(self, token: str = None, server: str = None, skip_tls: bool = False
...
+                    file.write(line)
+        print("logged out of user %s" % self.username)
 
 
-class PasswordUserAuthentication(Authentication):
+class KubeConfigFileAuthentication(KubeConfiguration):
     """
-    `PasswordUserAuthentication` is a subclass of `Authentication`. It can be used to authenticate to an OpenShift
-    cluster when the user has a username and password.
+    An abstract class that defines the necessary methods for passing a user's own Kubernetes config file.
```
MichaelClifford, line 252:
```suggestion
    A class that defines the necessary methods for passing a user's own Kubernetes config file.
```
I don't think this is an abstract class as its a subclass of KubeConfiguraiton.

Bobbins228:
I have changed how the login works by using the `kubernetes API` to handle an authenticated user. This is independent of whether a user has a loaded config file already.

When a user logs out the default config file is loaded again.

src/codeflare_sdk/cluster/auth.py
```diff
@@ -43,80 +48,126 @@ def logout(self):
...
-        self.password = password
-
-    def login(self) -> str:
-        """
-        This function is used to login to an OpenShift cluster using the user's `username` and `password`.
-        """
-        response = oc.login(self.username, self.password)
-        return response.out()
+    def __init__(self, kube_config_path: str = None):
+        self.kube_config_path = kube_config_path
```
MichaelClifford, line 147: Can this throw an error  or warning if `kube_config_path`  is left as `None`?

src/codeflare_sdk/cluster/auth.py
```diff
@@ -43,80 +48,126 @@ def logout(self):
...
+        self.ca_cert_path = ca_cert_path
 
     def login(self) -> str:
         """
-        This function is used to login to an OpenShift cluster using the user's API token and API server address.
-        Depending on the cluster, a user can choose to login in with "--insecure-skip-tls-verify` by setting `skip_tls`
-        to `True`.
+        This function is used to log in to a Kubernetes cluster using the user's API token and API server address.
+        Depending on the cluster, a user can choose to login in with `--insecure-skip-tls-verify` by setting `skip_tls`
+        to `True` or `--certificate-authority` by setting `skip_tls` to false and providing a path to a ca bundle with `ca_cert_path`.
```
Maxusmusti, line 102:
```suggestion
        to `True` or `--certificate-authority` by setting `skip_tls` to `False` and providing a path to a ca bundle with `ca_cert_path`.
```

src/codeflare_sdk/cluster/auth.py
```diff
@@ -43,80 +48,126 @@ def logout(self):
...
-                return error_msg
-        return response.out()
+                configuration.verify_ssl = False
+            api_client = client.ApiClient(configuration)
+            path_set = False
+            return "Logged into %s" % self.server
+        except client.ApiException as exception:
+            return exception
+
+    def api_config_handler():
```
Maxusmusti, line 121: Add function return type hint

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -114,8 +114,10 @@ def up(self):
         """
         namespace = self.config.namespace
         try:
-            config.load_kube_config()
-            api_instance = client.CustomObjectsApi()
+            KubeConfigFileAuthentication.config_check()
+            api_instance = client.CustomObjectsApi(
+                TokenAuthentication.api_config_handler()
```
Maxusmusti, line 119: Should this still be passed in if the config_check is successful?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -327,7 +333,7 @@ def list_all_queued(namespace: str, print_to_console: bool = True):
 
 def get_current_namespace():  # pragma: no cover
```
Maxusmusti, line 334: Note that this function will need to work regardless of auth method

Maxusmusti (changes requested):
Looking good, just a few questions and nits to start

src/codeflare_sdk/cluster/cluster.py
Bobbins228: Yes because if `config_check` passes the result of `api_config_handler()` will be `None` which is the default functionality if a user has a loaded config file.

src/codeflare_sdk/cluster/cluster.py
Bobbins228: I have altered the `config_check` function so that it will now pass the path to the user's specified config file. This will make it so that when `get_current_namespaces()` is called `config.list_kube_config_contexts()` will use the context from the user-defined config file.

With regards to when people authenticate using token and server.
`config.list_kube_config_contexts()` will not work because the user does not have a config file. 
However, when a user authenticates via this method a file is updated with the current namespace `/var/run/secrets/kubernetes.io/serviceaccount/namespace`.

I can add some functionality to read the current namespace from this file if it exists and if it doesn't I could set the namespace to `default` or throw an error to tell the user to specify a namespace themselves.

carsonmh:
Tested this out and works as expected. LGTM!

carsonmh (approved):

src/codeflare_sdk/cluster/cluster.py
Maxusmusti: Yes, I think adding something like that would be perfect. If we have any way of telling the user the current namespace being assumed, we should. It really mainly matters because when a `ClusterConfiguration` is created without a specified namespace, it will call `get_current_namespace` to fill that in for the user. So we also need to have appropriate error handling for the case where a user authenticates via server/token, defines a cluster without passing in namespace, and now we either have to gracefully tell them to set it manually (if we cant find it anywhere), or pick a default (maybe `default`). @MichaelClifford Any preference between default being `default` if no way to get namespace, vs printing a warning and returning `None` for `get_current_namespace`, and then a ValueError or something for the `ClusterConfiguration`?

MichaelClifford: Defaulting to "default" I do not think is a real option as the average user likely doesn't actually have permissions to the default namespace. 

When you authenticate to OpenShift you always get associated with some namespace at login. So calling `get_current_namespace()` is always an option. Does kube authentication allow you to login to a cluster, but not a specific namespace?   

If so, I think we should add some error handling to inform the user that they need to specify a namespace.

Maxusmusti (changes requested):
Quick issue I noticed when testing. This is the current behavior when namespace cannot be found:
 - `get_current_namespace()` returns: "Unable to find current namespace please specify with namespace=<your_current_namespace>"
 - `ClusterConfiguration()` (without setting namespace) works and generates yaml when passed into cluster with `namespace=Unable to find current namespace please specify with namespace=<your_current_namespace>`
 
 What we want is probably something more along the lines of:
  - `get_current_namespace()` returns `None` in that case and prints a message like "Unable to find current namespace"
  - `Cluster(ClusterConfiguration()` fails if the namespace in the configuration is set to `None`, and tells the user "Unable to find current namespace, please specify in your ClusterConfiguration with namespace=<your_current_namespace>"

Maxusmusti:
So like after line 71 in `cluster.py` just check if namespace is still `None`, and if so then fail with that message

MichaelClifford (changes requested):
Overall looks very good :+1: 

One thing though that is something of a simple change but, will impact a lot of the implementation. :grimacing: 

Can we make `api_config_handler()` and `config_check()` generic auth functions instead of class methods? It looks odd to invoke both authentication type classes on every call, especially when a users should only ever be using one anyways.

Is there any reason we can't separate these from the classes?

Maxusmusti (changes requested):
Everything seemed to work great, except one critical edge case:
 - When doing a token login, I can just pass in no token at all (or no server), and it leads to all sorts of broken behavior (those should not have defaults)
 - Also, when I pass in an incorrect token, it will still tell me I've "logged in", only to fail later down the line

Also, how are you obtaining/generating the ca cert that you are passing in when testing the Token approach? I'd like to test that as well, have just been testing with tls-skip for now.

EDIT: Realized if you're on RHEL the default path is probably where your ca cert bundle is, not sure how to do it on mac os though...

Bobbins228:
> Everything seemed to work great, except one critical edge case:
> 
> * When doing a token login, I can just pass in no token at all (or no server), and it leads to all sorts of broken behavior (those should not have defaults)
> * Also, when I pass in an incorrect token, it will still tell me I've "logged in", only to fail later down the line
> 
> Also, how are you obtaining/generating the ca cert that you are passing in when testing the Token approach? I'd like to test that as well, have just been testing with tls-skip for now.
> 
> EDIT: Realized if you're on RHEL the default path is probably where your ca cert bundle is, not sure how to do it on mac os though...

That was a great catch about default values for the token. I'll update the PR.
Yeah the default path is where the ca cert bundle is for RHEL and Openshift.

Bobbins228:
Looking into how we can pass an incorrect token and catch that and there doesn't seem to be any errors for when you set up the configuration even with bad credentials. Only when the api instance is used will an error be thrown e.g.`cluster.up()`. 

I found [this](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/AuthenticationApi.md), we can call `get_api_group()` to test if the credentials are correct and add in an error if that fails.

Maxusmusti:
@Bobbins228 the `get_api_group` solution seems great to me!

Maxusmusti (approved):
LGTM

Issue #183, opened 2023-06-27
Ray Cluster Retrieval Function


Maxusmusti:
Add a function to retrieve ray cluster objects from K8s/OCP, so that people can pick up and resume managing the cluster in fire-and-forget or pipeline scenarios.

Maxusmusti:
Some implementation details:

In the `cluster/cluster.py` file, we need a new function:
```
cluster_retrieval(cluster_name: str, namespace: str = "default") -> Cluster:
    This function should search the k8s/OCP cluster using the specified namespace for a RayCluster with the passed-in 
    name. If found, populate and return a Cluster object with all expected fields (via a ClusterConfig). If not, either return
    None or raise some form of "not found" error or exception.
```

Note that cluster_retrieval is a place holder name, and if anything better is come up with, it can be used instead.

MichaelClifford:
@Maxusmusti is there any issue with using the name `get_cluster(cluster_name)` instead?

Maxusmusti:
@MichaelClifford Nope, already a better suggestion 😅

PR #189, opened 2023-06-30
Get cluster

carsonmh:
Feature #183 

Makes function that returns Cluster object based on the cluster name and namespace
Tested using unit tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -335,53 +335,67 @@ def get_current_namespace():  # pragma: no cover
         return active_context["context"]["namespace"]
     except KeyError:
         return "default"
-    
+
 
 def get_cluster(cluster_name: str, namespace: str = "default"):
-   try:
-       config.load_kube_config()
-       api_instance = client.CustomObjectsApi()
-       rcs = api_instance.list_namespaced_custom_object(
-           group="ray.io",
-           version="v1alpha1",
-           namespace=namespace,
-           plural="rayclusters",
-       )
-   except Exception as e:
-       return _kube_api_error_handling(e)
+    try:
+        config.load_kube_config()
+        api_instance = client.CustomObjectsApi()
+        rcs = api_instance.list_namespaced_custom_object(
+            group="ray.io",
+            version="v1alpha1",
+            namespace=namespace,
+            plural="rayclusters",
+        )
+    except Exception as e:
+        return _kube_api_error_handling(e)
 
-
-   for rc in rcs["items"]:
-       if rc["metadata"]["name"] == cluster_name:
-           machine_types = rc["metadata"]["labels"]["orderedinstance"].split("_") if "orderedinstance" in rc["metadata"]["labels"] else []
-           local_interactive = "volumeMounts" in rc["spec"]["workerGroupSpecs"][0]["template"]["spec"]["containers"][0]
-           cluster_config = ClusterConfiguration(
-               name=rc["metadata"]["name"],
-               namespace=rc["metadata"]["namespace"],
-               machine_types=machine_types,
-               min_cpus=rc["spec"]["workerGroupSpecs"][0]["template"]["spec"]["containers"][
-                   0
-               ]["resources"]["requests"]["cpu"],
-               max_cpus=rc["spec"]["workerGroupSpecs"][0]["template"]["spec"]["containers"][
-                   0
-               ]["resources"]["limits"]["cpu"],
-               min_memory=int(rc["spec"]["workerGroupSpecs"][0]["template"]["spec"]["containers"][
-                   0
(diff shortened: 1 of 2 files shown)


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -337,6 +337,67 @@ def get_current_namespace():  # pragma: no cover
         return "default"
 
 
+def get_cluster(cluster_name: str, namespace: str = "default"):
+    try:
+        config.load_kube_config()
+        api_instance = client.CustomObjectsApi()
+        rcs = api_instance.list_namespaced_custom_object(
```
KPostOffice, line 344: https://github.com/kubernetes-client/python/blob/045b189a282f5165426b841e9e3b16b2739d3837/kubernetes/client/api/custom_objects_api.py#L1580

Can you use this directly?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -337,6 +337,67 @@ def get_current_namespace():  # pragma: no cover
...
+                    0
+                ]["resources"]["limits"]["nvidia.com/gpu"],
+                instascale=True if machine_types else False,
+                image=rc["spec"]["workerGroupSpecs"][0]["template"]["spec"][
+                    "containers"
+                ][0]["image"],
+                local_interactive=local_interactive,
+            )
+            return Cluster(cluster_config)
+    raise FileNotFoundError(
```
KPostOffice, line 396: If you are able to use the function suggested above then the error handling can be removed here.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -337,6 +337,67 @@ def get_current_namespace():  # pragma: no cover
...
+                gpu=rc["spec"]["workerGroupSpecs"][0]["template"]["spec"]["containers"][
+                    0
+                ]["resources"]["limits"]["nvidia.com/gpu"],
+                instascale=True if machine_types else False,
+                image=rc["spec"]["workerGroupSpecs"][0]["template"]["spec"][
+                    "containers"
+                ][0]["image"],
+                local_interactive=local_interactive,
+            )
+            return Cluster(cluster_config)
```
KPostOffice, line 395: This should be broken out into it's own function in order to improve readability.

Something like:
```python
def k8_cluster_obj_2_sdk_cluster_obj(rc):
```

KPostOffice (commented):
Looks good. Just a couple of comments.

src/codeflare_sdk/cluster/cluster.py
KPostOffice: maybe even a function in `Cluster` like `from_k8_cluster_object(rc)`

src/codeflare_sdk/cluster/cluster.py
carsonmh: I get `urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=80): Max retries exceeded with url: /apis/ray.io/v1alpha1/namespaces/default/rayclusters/quicktest` when calling the `get_namespaced_custom_object` so I think there is something wrong with this function

jbusche:
I tried it out on my OC 4.12 cluster and it's looked good - but Carson noticed a min/max workers aren't matching what was defined:
cluster.details() confirms 2/2 workers
```
 ╭────────────────────────────────────────────────────────────────╮ 
 │   Name                                                         │ 
 │   mnisttest                                        Active ✅   │ 
 │                                                                │ 
 │   URI: ray://mnisttest-head-svc.default.svc:10001              │ 
 │                                                                │ 
 │   Dashboard🔗                                                  │ 
 │                                                                │ 
 │                      Cluster Resources                         │ 
 │   ╭─ Workers ──╮  ╭───────── Worker specs(each) ─────────╮     │ 
 │   │  Min  Max  │  │  Memory      CPU         GPU         │     │ 
 │   │            │  │                                      │     │ 
 │   │  2    2    │  │  2~2         2           0           │     │ 
 │   │            │  │                                      │     │ 
 │   ╰────────────╯  ╰──────────────────────────────────────╯     │ 
 ╰────────────────────────────────────────────────────────────────╯
```
But the cluster.config is reporting just 1 each:
```
ClusterConfiguration(name='mnisttest', namespace='default', head_info=[], machine_types=[], min_cpus=2, max_cpus=2, min_worker=1, max_worker=1, min_memory=2, max_memory=2, gpu=0, template='/opt/app-root/src/codeflare-sdk/src/codeflare_sdk/templates/base-template.yaml', instascale=False, envs={}, image='ghcr.io/foundation-model-stack/base:ray2.1.0-py38-gpu-pytorch1.12.0cu116-20221213-193103', local_interactive=False)
```

jbusche (approved):
LGTM
Testing on my cluster, it looked good.  Thanks Carson

PR #244, opened 2023-07-25
Added conditions before deleting template elements

### LINKED ISSUE #206: Check for Template elements before deleting them in generate_yaml.py
We need to support the scenario where a user uses their own template yaml files to define their appwrappers.  Due to this scenario, we should not presume the existence of any fields in `generate_yaml.py` and should check for their existence before deleting them. For example,  `disable_raycluster_tls(resources)` will fail if a custom template does not define a volume. 

https://github.com/project-codeflare/codeflare-sdk/blob/ec6c787c6bc060fccb82f9b6e0684d3dfc12e016/src/codeflare_sdk/utils/generate_yaml.py#L263C1-L283

Bobbins228:
closes #206

### DIFF
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -302,14 +302,25 @@ def disable_raycluster_tls(resources):
             ]
         )
     ):
-        if i == 1:
+        if (
+            generic_template_spec["workerGroupSpecs"][0]["template"]["spec"][
+                "initContainers"
+            ][i]["name"]
+            == "create-cert"
+        ):
             del generic_template_spec["workerGroupSpecs"][0]["template"]["spec"][
                 "initContainers"
             ][i]
 
-    if len(resources["GenericItems"]) >= 3:
-        del resources["GenericItems"][3]  # rayclient route
-        del resources["GenericItems"][2]  # ca-secret
+    updated_items = []
+    for i in resources["GenericItems"][:]:
+        if "rayclient-deployment-name" in i["generictemplate"]["metadata"]["name"]:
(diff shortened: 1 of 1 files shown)


MichaelClifford (changes requested):
Looks good, but one issue with the way the last couple checks are done.

src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -302,14 +302,25 @@ def disable_raycluster_tls(resources):
...
-        del resources["GenericItems"][2]  # ca-secret
+    updated_items = []
+    for i in resources["GenericItems"][:]:
+        if "rayclient-deployment-name" in i["generictemplate"]["metadata"]["name"]:
+            continue
+        if "ca-secret-deployment-name" in i["generictemplate"]["metadata"]["name"]:
+            continue
+        updated_items.append(i)
+
+    resources["GenericItems"] = updated_items
```
MichaelClifford, line 323: Why aren't you deleting these items? @Bobbins228

Bobbins228: It's essentially the same right? 
`resources["GenericItems"]` is equated to `resources["GenericItems"]` without the ca-secret and ray-client. 

I tested this out and compared the yaml file with one generated with the current sdk and there were no differences. 

I could do what I did [here](https://github.com/project-codeflare/codeflare-sdk/pull/244/files#diff-279fd895cfcb8295f1b610826f8eb7a40fd543ff873661acd9dc452cef3d9d14R305-R313) if necessary.

MichaelClifford (approved):
LGTM!

PR #252, opened 2023-07-26
CLI Authentication

### LINKED ISSUE #217: CLI Cluster Authentication
Make CLI functionality allowing user to authenticate into their Kubernetes cluster. This will include login and logout functions.

carsonmh:
# Issue link
closes #217

# What changes have been made
Made login and logout functions in the CLI, allowing the user to authenticate into their Kubernetes cluster using token + server. Also added `load_auth` function which loads the authentication configuration so the SDK can use it.

# Verification steps
- `codeflare login` and `codeflare logout` are the two functions to be tested. Can be run in the terminal after building SDK with `pip install -e .`
- Unit tests can be run in usual way with `pytest -v tests/unit_test.py` or `python -m pytest tests/unit_test.py`

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cli/cli_utils.py
+++ b/src/codeflare_sdk/cli/cli_utils.py
@@ -50,7 +50,8 @@ def load_auth():
             auth = pickle.load(file)
             sdk_auth.api_client = auth.create_client()
             sdk_auth.config_path = auth.k8_config_path
+            return auth
     except IOError:
-        raise "Not logged into Kubernetes cluster"
+        return None
     except EOFError:
-        raise "Not logged into Kubernetes cluster"
+        return None
--- a/src/codeflare_sdk/cli/cli_utils.py
+++ b/src/codeflare_sdk/cli/cli_utils.py
@@ -1,5 +1,10 @@
 import ast
 import click
+from kubernetes import client
+import pickle
+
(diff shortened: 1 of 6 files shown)


Maxusmusti:
This would be a good one for @Bobbins228 to review as well, as he's pretty familiar with the new auth updates

Bobbins228 (approved):
LGTM! Can confirm login and logout work as they should and tests pass nice work Carson

Maxusmusti (changes requested):
Put this in slack, but adding this here as well so nobody accidentally merges without seeing:

Looks good, just a couple of things:
 - When things are saved in "auth", that's an actual file getting created and stored, right? If so, be careful about using a relative path, as people may login, move to a different dir, then be confused as to why they are logged out. We may instead want this file to be in a set `~/.codeflare/auth`
 - In your unit tests, it may be worth checking that the contents of the `auth` file are as you expect as well
 - Just checking, where is the `load_auth()` function going to be used, is it going to be at the start of every other CLI function?
 - I'm assuming this is just for token auth? That makes sense, as users can otherwise just do their own `export KUBECONFIG=...` if they need to change default kubeconfig path. I would just test and make sure that that option still works as expected

Maxusmusti (approved):
LGTM

PR #293, opened 2023-08-08
View and delete jobs functionality in the CLI

### LINKED ISSUE #277: CLI job view + delete

carsonmh:
# Issue link
closes #277 

# What changes have been made
Added 
- job status command
- job cancel command
- job logs command
- job list command
- `list_jobs` functionality in `cli_utils`
- `get_job` functionality in `cli_utils`

# Verification steps
build SDK with `pip install -e .` and run any command with `codeflare`.

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


src/codeflare_sdk/cli/cli_utils.py
```diff
@@ -73,3 +79,95 @@ def resolve_command(self, ctx, args):
...
+            "Job ID": job.job_id,
+            "RayCluster": cluster.name,
+            "Namespace": cluster.namespace,
+            "Status": str(job.status),
+            "App Handle": "ray://torchx/" + cluster.dashboard + "-" + job.submission_id,
+        }
+        rc_jobs.append(job_obj)
+    if print_to_console:
+        print_jobs(rc_jobs)
+    return rc_jobs
```
MichaelClifford, line 145: There is already a function for this in the cluster class `list_jobs()` can that be reused here instead?

MichaelClifford (changes requested):

src/codeflare_sdk/cli/cli_utils.py
carsonmh: @MichaelClifford Cluster's `list_jobs()` function returns jobs in a format that doesn't work with my table and doesn't include things like app handle. So this function is mainly reformatting the jobs so they work with a table and getting extra data.

MichaelClifford: Ok. We may want to look into some refactoring in the near future then, so that we are only maintaining a single function that gets_jobs info. But this is fine for now.

src/codeflare_sdk/cli/commands/list.py
```diff
@@ -16,15 +20,32 @@ def cli():
...
+
+
+@cli.command()
+@click.pass_context
+@click.option("--cluster-name", "-c", type=str)
+@click.option("--namespace", "-n", type=str)
+@click.option("--no-ray", is_flag=True)
+def job(ctx, cluster_name, namespace, no_ray):
+    """
+    List all jobs in a specified RayCluster or in K8S cluster
```
MichaelClifford, line 42: This is not exactly right. in both cases the jobs are being run on K8s clusters. However, they may be run either through the ray or mcad-kubernetes schedulers. In the ray case there will be a "raycluster" that we can reference to get job info. In the mcad-kubernetes case there is only pods that get deployed (as their own short lived "cluster"). 

So you may want to change this to just "List all jobs submitted".

src/codeflare_sdk/cli/commands/list.py
```diff
@@ -16,15 +20,32 @@ def cli():
...
         list_all_clusters(namespace)
         return
     list_clusters_all_namespaces()
+
+
+@cli.command()
+@click.pass_context
+@click.option("--cluster-name", "-c", type=str)
+@click.option("--namespace", "-n", type=str)
+@click.option("--no-ray", is_flag=True)
```
MichaelClifford, line 39: How about this is `--kube-mcad-scheduler-only`

MichaelClifford (changes requested):

MichaelClifford (approved):

MichaelClifford:
@carsonmh Can you rebase this PR?

PR #295, opened 2023-08-08
add and implement option to not generate appwrapper in a Cluster

### LINKED ISSUE #278: CLI (and SDK) Unnecessary yaml rewrites
Currently, when calling using the CLI's `status` or `details` commands, they will also write the cluster's AW back to current dir, since `get_cluster()` recreates the `Cluster` object. Perhaps we should have a `Cluster` creation option that doesn't actually generate the AW (just stores values in cluster/config object memory)? That way it can get consumed by `get_cluster()` optionally, and the CLI can then use that.

carsonmh:
# Issue link
Closes #278 

# What changes have been made
Changed Cluster definition so it optionally doesn't create an appwrapper. Changed `Cluster.from_k8_cluster_object` so it doesn't create an appwrapper. So when `get_cluster` is called, no appwrapper is generated

# Verification steps
Use `get_cluster` and it won't generate an appwrapper. Use Cluster object normally and it will. 

## Checks
- [x]  I've made sure the tests are passing. 
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -120,8 +120,8 @@ class Cluster:
         the MCAD queue.
         """
         if self.app_wrapper_yaml is None:
-            print("Error putting up RayCluster: AppWrapper yaml not generated")
-            return
+            self.app_wrapper_yaml = self.create_app_wrapper()
+            self.app_wrapper_name = self.app_wrapper_yaml.split(".")[0]
         namespace = self.config.namespace
         try:
             config_check()
@@ -144,6 +144,9 @@ class Cluster:
         associated with the cluster.
         """
         namespace = self.config.namespace
+        if not self.config.name and not self.app_wrapper_name:
+            print("Error taking down cluster: missing name or AppWrapper")
+            return
         try:
             config_check()
             api_instance = client.CustomObjectsApi(api_config_handler())
@@ -152,7 +155,7 @@ class Cluster:
                 version="v1beta1",
                 namespace=namespace,
                 plural="appwrappers",
-                name=self.config.name,
+                name=self.app_wrapper_name or self.config.name,
             )
         except Exception as e:  # pragma: no cover
             return _kube_api_error_handling(e)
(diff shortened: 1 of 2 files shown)


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -115,6 +119,9 @@ def up(self):
         Applies the AppWrapper yaml, pushing the resource request onto
         the MCAD queue.
         """
+        if self.app_wrapper_yaml is None:
+            print("Error putting up RayCluster: AppWrapper yaml not generated")
```
MichaelClifford, line 123: Can we replace this error statement with:
```
self.app_wrapper_name = self.app_wrapper_yaml.split(".")[0]
```

This way we just write it if it does not exists?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -145,7 +152,7 @@ def down(self):
                 version="v1beta1",
                 namespace=namespace,
                 plural="appwrappers",
-                name=self.app_wrapper_name,
+                name=self.config.name,
```
MichaelClifford, line 155: Are `self.app_wrapper_name` and `self.config.name` always the same? Is it safe to make this substitution?

MichaelClifford (changes requested):

src/codeflare_sdk/cluster/cluster.py
carsonmh: Good catch @MichaelClifford, the user could leave the name blank. Added a fix for this.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -137,6 +144,9 @@ def down(self):
         associated with the cluster.
         """
         namespace = self.config.namespace
+        if self.app_wrapper_yaml is None:
+            self.app_wrapper_yaml = self.create_app_wrapper()
```
MichaelClifford, line 148: Is there another way to get `self.app_wrapper_name` here other than generating the appwrapper yaml. Seems unnecessary to generate the appwrapper yaml just to delete the resource.

MichaelClifford (approved):

Issue #174, opened 2023-06-22
OAuth Proxy for Ray Dashboard


KPostOffice:
Right now ray dashboard is unauthenticated. Anyone with a URL can access it. We should consider adding an option for putting an OAuth proxy in front to provide auth.

https://github.com/oauth2-proxy/oauth2-proxy

anishasthana:
Giving this an 8 for now as it will be blocked by 3 issues to come

PR #298, opened 2023-08-10
add functions for creating ray with oauth proxy in front of the dashboard

KPostOffice:
# Issue link
closes: #174 

# What changes have been made
I added functions which create the necessary objects and update the appwrapper so that the SDK can create a ray cluster with an OAuth Proxy in front of the dashboard.

- OAuth sidecar container and tls secret mount
- Ingress with annotations for tls passthrough
- Service which points to OAuth port in the sidecar container and annotations for autogenerated tls certs
- Service account which contains the OAuth Callback
- ClusterRoleBinding giving OAuth sidecar necessary permissions

# Verification steps
go through notebook but set `openshift_oauth` to true in the config

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -21,12 +21,20 @@ cluster setup queue, a list of all existing clusters, and the user's working nam
 from time import sleep
 from typing import List, Optional, Tuple, Dict
 
+import openshift as oc
+from kubernetes import config
 from ray.job_submission import JobSubmissionClient
+import urllib3
 
 from .auth import config_check, api_config_handler
 from ..utils import pretty_print
 from ..utils.generate_yaml import generate_appwrapper
 from ..utils.kube_api_helpers import _kube_api_error_handling
+from ..utils.openshift_oauth import (
+    create_openshift_oauth_objects,
+    delete_openshift_oauth_objects,
+    download_tls_cert,
+)
 from .config import ClusterConfiguration
 from .model import (
     AppWrapper,
@@ -39,6 +47,8 @@ from kubernetes import client, config
 import yaml
 import os
 import requests
+
+from kubernetes import config
 
 
 class Cluster:
@@ -61,6 +71,39 @@ class Cluster:
         self.config = config
         self.app_wrapper_yaml = self.create_app_wrapper()
         self.app_wrapper_name = self.app_wrapper_yaml.split(".")[0]
+        self._client = None
+
+    @property
+    def _client_headers(self):
+        k8_client = api_config_handler() or client.ApiClient()
+        return {
+            "Authorization": k8_client.configuration.get_api_key_with_prefix(
+                "authorization"
+            )
+        }
+
+    @property
+    def _client_verify_tls(self):
+        return not self.config.openshift_oauth
+
+    @property
+    def client(self):
+        if self._client:
+            return self._client
+        if self.config.openshift_oauth:
+            print(
+                api_config_handler().configuration.get_api_key_with_prefix(
+                    "authorization"
+                )
+            )
+            self._client = JobSubmissionClient(
+                self.cluster_dashboard_uri(),
+                headers=self._client_headers,
+                verify=self._client_verify_tls,
+            )
+        else:
+            self._client = JobSubmissionClient(self.cluster_dashboard_uri())
+        return self._client
 
     def evaluate_dispatch_priority(self):
         priority_class = self.config.dispatch_priority
@@ -147,6 +190,7 @@ class Cluster:
             image_pull_secrets=image_pull_secrets,
             dispatch_priority=dispatch_priority,
             priority_val=priority_val,
+            openshift_oauth=self.config.openshift_oauth,
         )
 
     # creates a new cluster with the provided or default spec
@@ -156,6 +200,11 @@ class Cluster:
         the MCAD queue.
         """
         namespace = self.config.namespace
+        if self.config.openshift_oauth:
+            create_openshift_oauth_objects(
+                cluster_name=self.config.name, namespace=namespace
+            )
+
         try:
             config_check()
             api_instance = client.CustomObjectsApi(api_config_handler())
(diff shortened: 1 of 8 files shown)


KPostOffice:
This isn't quite finished yet. I'm having tls related issues between the Ray API and the OAuth proxy.

KPostOffice:
Blocked due to required changes upstream with the Ray Job Submission Client and TorchX

blublinsky:
This URL is used not only for dashboard access but also for the APIs - currently jobs and In future also  Serve. Oath proxy requires manual log in, which will break these APIs

KPostOffice:
@blublinsky I've launched OAuth so that it supports both GUI manual login as well as authentication using an auth bearer token. Can you PTAL?

blublinsky:
> @blublinsky I've launched OAuth so that it supports both GUI manual login as well as authentication using an auth bearer token. Can you PTAL?

How do you get the auth bearer token to use?

KPostOffice:
> > @blublinsky I've launched OAuth so that it supports both GUI manual login as well as authentication using an auth bearer token. Can you PTAL?
> 
> How do you get the auth bearer token to use?

https://github.com/project-codeflare/codeflare-sdk/pull/298/files#diff-5706a96c5346dd54a22370211260411363c4ed41d798fddf240eb8983b1366eaR82-R86

It is the same token used for interacting with the kubernetes api with the kubectl client. I grab it here based on the currently logged in user or in cluster config. I believe it is the same token as the one you generate from the `copy login command` page in the OpenShift client.

blublinsky:
> It is the same token used for interacting with the kubernetes api with the kubectl client. I grab it here based on the currently logged in user or in cluster config. I believe it is the same token as the one you generate from the `copy login command` page in the OpenShift client.

So basically the only thing that you are validating is that user is logged in. Not very secure, really

KPostOffice:
> > It is the same token used for interacting with the kubernetes api with the kubectl client. I grab it here based on the currently logged in user or in cluster config. I believe it is the same token as the one you generate from the `copy login command` page in the OpenShift client.
> 
> So basically the only thing that you are validating is that user is logged in. Not very secure, really

https://github.com/project-codeflare/codeflare-sdk/pull/298/files#diff-279fd895cfcb8295f1b610826f8eb7a40fd543ff873661acd9dc452cef3d9d14R423

The OAuth Proxy supports checking for authorization based on RBAC. Here I check to make sure the authenticated user has authorization to get pods in the given namespace.

blublinsky:
So it is basically a single cluster solution

KPostOffice:
> So it is basically a single cluster solution

I'm unsure what you are expecting. How would you propose we authenticate for a multi-cluster design here? AFAIK, the `codeflare-sdk` expects the user to be logged into a specific k8 cluster.

blublinsky:
> I'm unsure what you are expecting. How would you propose we authenticate for a multi-cluster design here? AFAIK, the `codeflare-sdk` expects the user to be logged into a specific k8 cluster.

I do not think authentication/authorization should be linked to a given cluster, It should rather be for a given user, who can be logged in any of the several clusters, but that's just my opinion

KPostOffice:
I was running into issues with the generate CA cert from the service annotation. Putting this here so I remember to look into the ca-injector.

https://cert-manager.io/docs/concepts/ca-injector/

KPostOffice:
> > I'm unsure what you are expecting. How would you propose we authenticate for a multi-cluster design here? AFAIK, the `codeflare-sdk` expects the user to be logged into a specific k8 cluster.
> 
> I do not think authentication/authorization should be linked to a given cluster, It should rather be for a given user, who can be logged in any of the several clusters, but that's just my opinion

Where do you propose authentication and authorization occur in this case?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -61,6 +69,26 @@ def __init__(self, config: ClusterConfiguration):
         self.config = config
         self.app_wrapper_yaml = self.create_app_wrapper()
         self.app_wrapper_name = self.app_wrapper_yaml.split(".")[0]
+        self._client = None
+        self._k8_client = config.new_client_from_config()
```
MichaelClifford, line 73: I think config is getting overloaded here. I just got the error: `AttributeError: 'ClusterConfiguration' object has no attribute 'new_client_from_config'`. 

It seems like you want to use `kubernetes.config` but just above you are setting `self.config` to the ClusterConfiguration object.

MichaelClifford (dismissed):

src/codeflare_sdk/cluster/cluster.py
KPostOffice: Thanks, nice catch

MichaelClifford:
Tested out starting a ray cluster from local machine and submitting jobs. Seems to work as expected when logging in via sdk authentication or `oc login`. 

lgtm

@Maxusmusti could you please review as well.

anishasthana:
@KPostOffice could you create a follow-on task to switch this over to istio (once we have istio available ootb in ODH). Or at least make it configurable somehow. Alternatively, the follow-on winds up being "Remove this, and update the kuberay operator to allow auth enablement"

src/codeflare_sdk/utils/kube_api_helpers.py
```diff
@@ -42,3 +43,7 @@ def _kube_api_error_handling(e: Exception):  # pragma: no cover
         elif e.reason == "Conflict":
             raise FileExistsError(exists_msg)
     raise e
+
+
+def _get_api_host(api_client: client.ApiClient):  # pragma: no cover
```
Bobbins228, line 48: When creating the appwrapper while authenticated with a K8s config file I experienced an error `'NoneType' object has no attribute 'configuration'`

When I ran through these changes while authenticated with token and server and it ran as expected. Just need to fix the _get_api_host function

Bobbins228 (changes requested):

Bobbins228 (approved):
Verified working on OSD cluster awesome work Kevin /approve

PR #302, opened 2023-08-11
fix get_current_namespace

### LINKED ISSUE #289: get_current_namespace broken due to authentication change
Currently `get_current_namespace` searches for a namespace file if there is an api_client present. The problem is, with the new authentication implementation, `oc login` isn't invoked, meaning that this namespace file may not be present. So its better to just use kubernetes API to search for namespace

carsonmh:
# Issue link
closes #289 

# What changes have been made
Sometimes when the user has an api_client, they don't always want to search the serviceaccount. So this PR changes the `get_current_namespace` so that it always checks both the serviceaccount and kubeconfig for current namespace

# Verification steps
Try logging out of cluster with `oc logout` and logging in through `TokenAuthentication`. Calling `get_current_namespace` should yield the correct result. With the old `get_current_namespace` it would print that namespace is not available

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -384,29 +384,23 @@ def list_all_queued(namespace: str, print_to_console: bool = True):
 
 
 def get_current_namespace():  # pragma: no cover
-    if api_config_handler() != None:
-        if os.path.isfile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"):
-            try:
-                file = open(
-                    "/var/run/secrets/kubernetes.io/serviceaccount/namespace", "r"
-                )
-                active_context = file.readline().strip("\n")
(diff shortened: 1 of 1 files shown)


Bobbins228 (changes requested):
When testing this on an OSD Cluster I was unable to delete `/var/run/secrets/kubernetes.io/serviceaccount/namespace` because it is read only. I changed the code to look for a file that doesn't exist instead.

When creating an appwrapper after logging in via the token authentication in a Jupyter Notebook I got a `ConfigException: Invalid kube-config file. No configuration found.` error.

As things stand at the moment if a user lacks a config file and the namespace file there is an error when calling `get_current_namespace()` even if they have authenticated via the token and server with `TokenAuthentication`.

Issue #274, opened 2023-08-03
Investigate use of poetry for dependency management


Fiona-Waters:
## WHY
We currently use both poetry and a requirements.txt file for managing dependencies. We want to investigate if complexity could be reduced by using poetry alone.

## WHAT
Investigate possible options for the sdk using poetry. Discuss with the team and implement.

## TESTS
Add tests as required.
## DONE
A decision has been made regarding the management of dependencies and any required changes have been made and tested.

Fiona-Waters:
Draft PR: https://github.com/project-codeflare/codeflare-sdk/pull/305/

PR #305, opened 2023-08-14
Refactoring to Use Poetry

Fiona-Waters:
# Issue link
#274 

# What changes have been made
 - The .gitignore has been updated so that the poetry.lock file is committed to the repo - allowing everyone to use the same versions of dependencies.
 - The main requirements.txt file in the root of the repo has been removed.
 - The readme has been updated to include `poetry install` rather than using pip
     - Instructions have also been added allowing creation of a requirements.txt file if required

 - The pre-commit Containerfile has been updated to use poetry, the build command will now be run from the root directory to allow access to the pyproject.toml file and all required dependencies.

 - Any installation commands and documentation/readmes have been updated to reflect the changes.

 - We will not remove the requirements.txt files from the demo notebooks. For more info see thread [here](https://project-codeflare.slack.com/archives/C04LK0G441F/p1692108534524669?thread_ts=1692010942.125799&cid=C04LK0G441F).

# Verification steps
- First checkout this pr
- Run `poetry install`
- Run `poetry build`

## To verify the building of the precommit image
 - In the root directory run `podman build -f .github/build/Containerfile .` and ensure that the build completes successfully.

## Ensure that unit tests pass
 - In the root directory run the unit tests with `poetry run pytest -v tests/unit_test.py`  

## Test demo notebooks
 - Create a cluster
 - Clone the distributed workloads repo
 - Log in to your cluster
 - Install the codeflare stack from the distributed workloads repo by running the following command `make all-in-one`
 - Once the installation of all components is complete, navigate to the ODH dashboard and to a Jupyter Notebook, and navigate to the terminal.
 - Uninstall the current codeflare-sdk by running  `pip uninstall codeflare-sdk -y`
 - Drag and drop the repo build created by the poetry build command run earlier into the Jupyter notebook and run `pip install codeflare_sdk-0.0.0.dev0-py3-none-any.whl`
- Restart the Notebook kernel
- Run through a couple of demo notebooks and ensure that they run successfully

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


.github/build/Containerfile
```diff
@@ -5,10 +5,20 @@ LABEL summary="Toolchain for running pre-commit hooks." \
...
-RUN python -m pip install -r /tmp/requirements-dev.txt
+RUN pip3 install poetry
+COPY pyproject.toml ./
+RUN poetry install
 
-CMD bash
+#COPY requirements.txt /tmp/requirements.txt
+#RUN python -m pip install -r /tmp/requirements.txt
+
+CMD bash
```
Maxusmusti, line 24: Missing newline at EOF

README.md
```diff
@@ -18,10 +18,12 @@ Can be installed via `pip`: `pip install codeflare-sdk`
 ### Prerequisites
 
 We recommend using Python 3.9 for development.
-Install development specific dependencies:
-  `$ pip install -r requirements-dev.txt`
+Install dependencies: `poetry install`
```
Maxusmusti, line 21: Should we include instructions for people who want to install just the base dependancies in the pyproject, vs just the test dependencies vs the docs dependancy?

Maxusmusti (changes requested):

.github/build/Containerfile
Maxusmusti: Also curious, why are the steps and requirements for the pre-commit image changing so much? It seems like before all we needed were the contents of `requirements-dev.txt`:
```
pre-commit
poetry
pytest
pytest-mock
coverage
black==22.3.0
```
but now we are doing a lot more than just those

Fiona-Waters: Thanks for the review @Maxusmusti 
In an attempt to test out using poetry instead of pip I copied the repos pyproject.toml file in here. If we want to go that way I can reduce it to just what is needed. 
The requirements-dev file was removed and its contents added to the requirements.txt, possibly to reduce the number of files - @anishasthana might be able to confirm this. The requirements file will be removed if we go with using the toml file instead.

README.md
Fiona-Waters: Yes we can do this. Currently the docs dependencies are optional so we can use `--with docs` to include them. Test dependencies are not currently optional so will be installed with the base dependencies. Do we want to make them optional? Or note to edit the file to make them optional if not required?

Maxusmusti: Yeah I think that having the test dependencies also be optional, but with a note on how to install those as well, might be the way to go. Because at the base, people just using the package from source should get just the required dependencies to get running by default. And then for developers, they should include the test dependencies themselves, run the `pre-commit install`, etc. As long as we are clear in our README about how to install test/doc dependencies, it should be fine

.github/build/Containerfile
Maxusmusti: @anishasthana thoughts? This is specifically for requirements for the pre-commit image. Also @Fiona-Waters would it be possible to pull these requirements in any way from the root dir pyproject, since they should exactly align with the test dependencies we have, that way we don't have to manage them in multiple places?

Fiona-Waters: I tried but wasn't able to pull them and use them through the Containerfile. Open to suggestions!

README.md
Fiona-Waters: Great, I will update this then. Thank you!

.github/build/Containerfile
anishasthana: In my opinion if we can have just one file to deal with, that'd be simplest.

Agreed on pulling from the root directory if possible. What if we ran the image build from the root directory? Would it give us access to all the stuff?

So like `podman build -f .github/build/Containerfile .` from the root path. I think it'll give you access to the complete context, so no copy needed.

Fiona-Waters: Thanks @anishasthana I've run this and it built successfully copying the pyproject.toml and resolving dependencies. Where should this change of building from root path be documented, other than in the readme?
We will also need to update the [coverage-badge.yaml](https://github.com/project-codeflare/codeflare-sdk/blob/4593dd5cd07c0a737150961302ae5ed76ed6da68/.github/workflows/coverage-badge.yaml#L26) to account for the removal of the requirements.txt file

astefanutti (commented):
It'd be great to update the SDK e2e test and make sure it passes with Poetry:

https://github.com/project-codeflare/codeflare-operator/blob/6f7d6b1194dd88aa0b2ab08b5b0e1856793883dd/test/e2e/mnist_raycluster_sdk_test.go#L164

Fiona-Waters:
> It'd be great to update the SDK e2e test and make sure it passes with Poetry:
> 
> https://github.com/project-codeflare/codeflare-operator/blob/6f7d6b1194dd88aa0b2ab08b5b0e1856793883dd/test/e2e/mnist_raycluster_sdk_test.go#L164

Thanks for the review @astefanutti Yes we should definitely do this. Do you think it would be okay if I create a new/follow on ticket for this, or would that need to be completed before merging?

astefanutti:
> > It'd be great to update the SDK e2e test and make sure it passes with Poetry:
> > https://github.com/project-codeflare/codeflare-operator/blob/6f7d6b1194dd88aa0b2ab08b5b0e1856793883dd/test/e2e/mnist_raycluster_sdk_test.go#L164
> 
> Thanks for the review @astefanutti Yes we should definitely do this. Do you think it would be okay if I create a new/follow on ticket for this, or would that need to be completed before merging?

@Fiona-Waters creating a follow up ticket to have it done along the next SDK dependency upgrade in the CodeFlare operator sounds the best course of action 👍🏼.

Fiona-Waters:
> Excellent job on this PR! 👏
> 
> To fix the pre-commit checks:
> 
> * Extra white spaces around the Readme file: if using VSCode what I like to do is activate Render Whitespace (View>Appearance>Render Whitespace). This shows all trailing white spaces in code that are not easily seen.
> * End-of-file-fixer: the hook checks that there should be an empty line at the very end of all files.

pre-commit now passing! Thanks @ChristianZaccaria

Fiona-Waters:
> > > It'd be great to update the SDK e2e test and make sure it passes with Poetry:
> > > https://github.com/project-codeflare/codeflare-operator/blob/6f7d6b1194dd88aa0b2ab08b5b0e1856793883dd/test/e2e/mnist_raycluster_sdk_test.go#L164
> > 
> > 
> > Thanks for the review @astefanutti Yes we should definitely do this. Do you think it would be okay if I create a new/follow on ticket for this, or would that need to be completed before merging?
> 
> @Fiona-Waters creating a follow up ticket to have it done along the next SDK dependency upgrade in the CodeFlare operator sounds the best course of action 👍🏼.

Ticket created - https://github.com/project-codeflare/codeflare-operator/issues/250

Maxusmusti (changes requested):
Ah I figured out why the unit tests are failing:
```
    import executing
E   ModuleNotFoundError: No module named 'executing'
```

In the `pyproject.toml`, `executing` should be a base dependency, not a test dependency. It's a package required in the SDK itself

Fiona-Waters:
> Ah I figured out why the unit tests are failing:
> 
> ```
>     import executing
> E   ModuleNotFoundError: No module named 'executing'
> ```
> 
> In the `pyproject.toml`, `executing` should be a base dependency, not a test dependency. It's a package required in the SDK itself

I was able to get the tests to pass by adding `pip install executing` [here](https://github.com/project-codeflare/codeflare-sdk/pull/305/files#diff-cd0effce06f425599e96952f4d5f684641a683f46d9909e7c74a9f8b2f03ccd3R23). This is more of a workaround than a fix I think. I am stumped otherwise as the module `executing` is listed in pyproject.toml and poetry.lock file. Please let me know what you think.

Maxusmusti (changes requested):
Worth noting, I get the following warning when doing a poetry install:
```
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
```

Also, I noticed the behavior that it *said* `executing` was being installed, but did not actually install the package. Similarly, however, I noticed that it also tries to install `codeflare-sdk==v0.0.0-dev`, but similarly does NOT actually install it...

It seems that this is not just a test issue, but actually a wider issue with the `poetry install`...

Upon further inspection, it seems as though `poetry install` isn't actually installing the packages to any default or desired python version, but to a poetry-specific virtual-env accessible via `poetry run ...` or `poetry shell`. See: https://python-poetry.org/docs/basic-usage/. This can be confirmed by prepending `poetry run` to the regular `pytest` command and it working. There seem to be ways to work around it (here's the first, kind of odd, one I found: https://medium.com/opsops/using-poetry-within-docker-container-without-venv-f8f07c9deba3), but not sure if we want to, or if we just want to update our usage patterns to reflect this discovered fact.

Fiona-Waters:
> Worth noting, I get the following warning when doing a poetry install:
> 
> ```
> Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
> ```
> 
> Also, I noticed the behavior that it _said_ `executing` was being installed, but did not actually install the package. Similarly, however, I noticed that it also tries to install `codeflare-sdk==v0.0.0-dev`, but similarly does NOT actually install it...
> 
> It seems that this is not just a test issue, but actually a wider issue with the `poetry install`...
> 
> Upon further inspection, it seems as though `poetry install` isn't actually installing the packages to any default or desired python version, but to a poetry-specific virtual-env accessible via `poetry run ...` or `poetry shell`. See: https://python-poetry.org/docs/basic-usage/. This can be confirmed by prepending `poetry run` to the regular `pytest` command and it working. There seem to be ways to work around it (here's the first, kind of odd, one I found: https://medium.com/opsops/using-poetry-within-docker-container-without-venv-f8f07c9deba3), but not sure if we want to, or if we just want to update our usage patterns to reflect this discovered fact.

Thanks so much for reviewing again @Maxusmusti 
So it looks like we have a few options: 
1. We use a poetry virtual env inside the container (which may not make sense).
2. We set `config virtualenvs.create false` and see if it works properly that way (I tried a few things but didn't get it right yet).
3. We decide not to go ahead with this refactor and use of poetry. 
I hope I understand this correctly.  I think maybe we try option 2 (not only because I've been working on it for a while! :) ) but I will defer to you and @anishasthana on this.

anishasthana:
I'm good with trying out `2.` first. @KPostOffice fyi

anishasthana:
>Would option 1 just be adding poetry run in front of the test command? Are there any concerns with doing that? I'm not opposed to either one or two.

Would this actually work for local dev? Or would we need to prepend "poetry run" to every command?

Maxusmusti:
@anishasthana yeah local dev is also affected by this. Developer flow would look like:
```
poetry install (+ any options)
```
and then either:
```
poetry shell
cmd
cmd
cmd
...
```
or
```
poetry run cmd
poetry run cmd
```
which isn't necessarily a bad thing, as it may be a good practice moving forward, but it definitely is a bit more limiting, and will affect some developers current setups/workflows

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  unit-tests:
+
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/project-codeflare/codeflare-sdk-precommit:v0.0.3
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
```
KPostOffice, line 17:
```yaml
      - name: Install dependencies
      - uses: actions/setup-python@v4
        with:
          python-version: '3.7'
          cache: 'poetry' # caching poetry dependencies
```

WDYT about caching dependencies here?

See: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

KPostOffice:
> I'm good with trying out `2.` first. @KPostOffice fyi

I think 2 is a bit better, but I have no real preference.

Fiona-Waters:
@anishasthana @Maxusmusti @KPostOffice 
I've updated the pre-commit unit-tests.yml to remove the poetry virtual env, and it is now passing. I had an error with `GLIBC` version so it's running with Python3.8. Please let me know if that is not okay.
I have updated the README.md to include `poetry run` for necessary commands. 
I think this is everything that is currently outstanding (apart from rebasing), please let me know if you see anything else. Thank you all so much for your help with this.

README.md
```diff
@@ -21,25 +21,38 @@ Can be installed via `pip`: `pip install codeflare-sdk`
 ### Prerequisites
 
 We recommend using Python 3.9 for development.
-Install development specific dependencies:
-  `$ pip install -r requirements-dev.txt`
 
-Additional dependencies can be found in `requirements.txt`: `$ pip install -r requirements.txt`
+#### Install dependencies:
+  - run: `poetry install`
```
anishasthana, line 26: I wonder if we should instead use `poetry shell` and go from there. Might be less annoying than repeatedly adding `poetry run`

Fiona-Waters: Updated. Thanks Anish.

Bobbins228 (approved):
/lgtm

.github/workflows/unit-tests.yml
anishasthana: @Fiona-Waters can you link the follow-up issue here?

Fiona-Waters: https://github.com/project-codeflare/codeflare-sdk/issues/345

KPostOffice (approved):

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install poetry
+        run: pip install poetry
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+      - name: Install dependencies
+        run: |
+          poetry config virtualenvs.create false
```
Maxusmusti, line 24: Does this line also need to be included for the pre-commit Containerfile and the coverage-badge workflow?

Maxusmusti (changes requested):
Looks good, just one question about other places where the virtual env may need to be turned off. Also, the poetry.lock file needs to be updated once more to update the `codeflare-torchx` dependency version

.github/workflows/unit-tests.yml
Fiona-Waters: Yes it should be, thanks for spotting that. Updated that and the poetry.lock file.

PR #306, opened 2023-08-14
SDK-e2e

### LINKED ISSUE #292: SDK e2e Testing
Pulling SDK testing in from `codeflare-operator` to be run on PRs and main branch of SDK repo.

Example e2e test for SDK from operator repo: https://github.com/project-codeflare/codeflare-operator/blob/main/test/e2e/mnist_raycluster_sdk_test.go

Distributed workloads tests example: https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/basictests/distributed-workloads.sh

ChristianZaccaria:
# Issue link

Closes #292 

# What changes have been made

- Added e2e test workflow.
- Added e2e test that creates a Ray cluster, and trains the MNIST dataset using the CodeFlare SDK. Asserts successful completion of the training job. Covering the installation of the CodeFlare SDK, as well as the RBAC required for the SDK to successfully perform requests to the cluster.
- Added action to setup KinD cluster and configuration.
- Added SDK script, MNIST training script, pip requirements, and SDK installation script.
- Initilialized go.mod and go.sum to make use of common codeflare functions.
- Created a script that copies the root directory (codeflare-sdk) to the volume mount in the Job container. This is done to install the codeflare-sdk with poetry with the latest changes, i.e., for opened PRs.

# Verification steps

To run the **e2e test locally on KinD cluster**, follow these steps:
- Setup Phase:
  - Pull the [codeflare-operator repo](https://github.com/project-codeflare/codeflare-operator) and run the following make targets:
  ```
  make kind-e2e
  export CLUSTER_HOSTNAME=kind
  export CLUSTER_TYPE=KIND
  export CODEFLARE_TEST_TIMEOUT_LONG=20m
  make deploy -e IMG=quay.io/project-codeflare/codeflare-operator:v1.0.0-rc.4
  make setup-e2e
  ```
- Test Phase:
   - Once we have the codeflare-operator and kuberay-operator running and ready, we can run the e2e test on the codeflare-sdk repo:
  ```
  go test -timeout 30m -v ./tests/e2e
  ```

To run the **e2e on HyperShift cluster**, follow these steps:
- Setup Phase:
  - Pull the [codeflare-operator repo](https://github.com/project-codeflare/codeflare-operator) and run the following make targets:
  ```
  export CLUSTER_TYPE=HYPERSHIFT
  export CODEFLARE_TEST_TIMEOUT_LONG=20m
  make deploy -e IMG=quay.io/project-codeflare/codeflare-operator:v1.0.0-rc.4
  make setup-e2e
  ```
- Test Phase:
   - Once we have the codeflare-operator and kuberay-operator running and ready, we can run the e2e test on the codeflare-sdk repo:
  ```
  go test -timeout 30m -v ./tests/e2e
  ```

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


tests/e2e/sdk-to-pod.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+echo "Starting the script to copy the SDK to the pod"
```
sutaakar, line 2: Is this script related to `tests/e2e/codeflare_sdk-0.0.0.dev0-py3-none-any.whl`?

sutaakar:
@ChristianZaccaria Can you describe how is the SDK installed into the job image?
Also thinking, would it be possible/have more sense to build image with SDK as part of the workflow, uploading it on KinD and then directly using the built image? That would skip all the complex handling in test itself, making it more readable.

tests/e2e/sdk-to-pod.sh
ChristianZaccaria: Hmm no, good catch I forgot to delete that wheel. In the installation of the codeflare-sdk for this test there is no wheel involved, so `poetry build` is never performed.

ChristianZaccaria:
> @ChristianZaccaria Can you describe how is the SDK installed into the job image? Also thinking, would it be possible/have more sense to build image with SDK as part of the workflow, uploading it on KinD and then directly using the built image? That would skip all the complex handling in test itself, making it more readable.

Great question! My understanding is that when we run the test, once the Job container is `running`, the container waits until the `pyproject.toml` file has been copied over and exists in the codeflare-sdk volume mount:
```
Command: []string{
	"/bin/sh", "-c",
	"while [ ! -f /codeflare-sdk/pyproject.toml ]; do sleep 1; done; " +
	"cp /test/* . && chmod +x install-codeflare-sdk.sh && ./install-codeflare-sdk.sh && python mnist_raycluster_sdk.py " + namespace.Name, },
```
While it is waiting for that condition to be true, what is happening in the background is that the function that executes the [`sdk-to-pod.sh` is ran concurrently](https://github.com/project-codeflare/codeflare-sdk/pull/306/files#diff-cc9eeef93126101c397a82a09a08aa1c51a4f77bf366a9db7176351fd81b3d7e:~:text=go%20func()%20%7B,%7D()). The sdk-to-pod script is provided with the generated namespace from the test, and attempts to read the pod name and status. Once the found pod has a status of `Running`, the root folder (all contents of local codeflare-sdk) are copied to the volume mount codeflare-sdk in the pod.

After we have our files in the volume mount, we do the same as originally done in the [test in CFO](https://github.com/project-codeflare/codeflare-operator/blob/main/test/e2e/mnist_raycluster_sdk_test.go#L124), we install the SDK with the `install-codeflare-sdk.sh`.

This was done this way because, originally in the test in CFO repo the SDK is installed by performing a git clone against main branch of the SDK, then installing it in the codeflare-sdk volume mount by using the mentioned script, and I was looking to maintain a similar behaviour.

Let me know if this makes sense, thanks Karel!

tests/e2e/mnist_raycluster_sdk_test.go
```diff
@@ -0,0 +1,236 @@
...
+							// FIXME: switch to base Python image once the dependency on OpenShift CLI is removed
+							// See https://github.com/project-codeflare/codeflare-sdk/pull/146
+							Image: "quay.io/opendatahub/notebooks:jupyter-minimal-ubi8-python-3.8-4c8f26e",
+							Env: []corev1.EnvVar{
+								{Name: "PYTHONUSERBASE", Value: "/workdir"},
+								{Name: "RAY_IMAGE", Value: GetRayImage()},
+							},
+							Command: []string{
+								"/bin/sh", "-c",
+								"while [ ! -f /codeflare-sdk/pyproject.toml ]; do sleep 1; done; " +
```
astefanutti, line 152: If we stick to using that approach (vs. the test container image build one), I'd suggest we move these commands in an init container.

tests/e2e/mnist_raycluster_sdk_test.go
```diff
@@ -0,0 +1,236 @@
...
+		hostnameEnvVar := corev1.EnvVar{Name: "CLUSTER_HOSTNAME", Value: hostname}
+		job.Spec.Template.Spec.Containers[0].Env = append(job.Spec.Template.Spec.Containers[0].Env, hostnameEnvVar)
+	}
+	job, err := test.Client().Core().BatchV1().Jobs(namespace.Name).Create(test.Ctx(), job, metav1.CreateOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+	test.T().Logf("Created Job %s/%s successfully", job.Namespace, job.Name)
+
+	go func() {
+        scriptName := "./sdk-to-pod.sh"
+        cmd := exec.Command(scriptName, namespace.Name)
```
astefanutti, line 212: I'd suggest to use the `k8s.io/kubectl/pkg/cmd/cp` package directly, e.g.:

```go
import "k8s.io/kubectl/pkg/cmd/cp"

streams, in, out, err := genericclioptions.NewTestIOStreams()
copy := cp.NewCopyOptions(streams)
...
err := copy.Run([]string{src, dst})
```

tests/e2e/install-codeflare-sdk.sh
```diff
@@ -0,0 +1,17 @@
...
+cd ..
+
+# Install Poetry and configure virtualenvs
+pip install poetry
+poetry config virtualenvs.create false
+
+cd codeflare-sdk
+
+# Lock dependencies and install them
+poetry lock --no-update
```
Bobbins228, line 12: I am wondering why we run `poetry lock --no-update` if the `poetry.lock` file should already exist?

Bobbins228 (commented):
Ran the test on Kind with no issue just have to address Antonin's comment and this PR should be good to go.

tests/e2e/install-codeflare-sdk.sh
ChristianZaccaria: When we run `poetry lock --no-update` we are resolving the dependencies from the project.toml file, which then creates the `poetry.lock` file. You're right, the poetry.lock file already exists, but if any changes are made to the project.toml file we would need to update our poetry.lock file too. The mentioned command ensures project.toml and poetry.lock are in sync.

sutaakar:
@ChristianZaccaria Have you considered whether it has more sense for e2e test to be implemented in Python instead of Go, to install and leverage SDK directly?

ChristianZaccaria:
> @ChristianZaccaria Have you considered whether it has more sense for e2e test to be implemented in Python instead of Go, to install and leverage SDK directly?

I think @astefanutti suggested we kept it as a go test, but this could change. 

We were discussing that in this PR we currently have a well working e2e test and that it might be best to merge it to gate PRs and unblock dependabot updates. 

To not diverge from the purpose of this PR, we have this issue opened to refactor/simplify the test and make use of unprivileged users: https://issues.redhat.com/browse/RHOAIENG-60. In the scope of that issue most items including the job spec, RBAC, and configmap in go test will be removed. Taking that into account as well, you're right I think we should implement the e2e test all in Python.

I'll be working on the new issue with @Bobbins228 if you think that's okay :)

tests/e2e/mnist_raycluster_sdk_test.go
ChristianZaccaria: As discussed previously, your suggested solution is from older versions of kubectl where the signature changed quite a lot. I made the necessary changes to implement the `cp` package directly now instead of using the script. Thank you!

tests/e2e/mnist_raycluster_sdk_test.go
ChristianZaccaria: This will be addressed as part of refactoring/simplifying the test and make use of unprivileged users: https://issues.redhat.com/browse/RHOAIENG-60

sutaakar:
Sure, fine for me

.github/workflows/e2e_tests.yaml
```diff
@@ -0,0 +1,127 @@
...
+        run: |
+          echo "Printing CodeFlare operator logs"
+          kubectl logs -n openshift-operators --tail -1 -l app.kubernetes.io/name=codeflare-operator | tee ${CODEFLARE_TEST_OUTPUT_DIR}/codeflare-operator.log
+
+      - name: Print KubeRay operator logs
+        if: always() && steps.deploy.outcome == 'success'
+        run: |
+          echo "Printing KubeRay operator logs"
+          kubectl logs -n ray-system --tail -1 -l app.kubernetes.io/name=kuberay | tee ${CODEFLARE_TEST_OUTPUT_DIR}/kuberay.log
+
```
sutaakar, line 119: A nitpick, you can export all KinD pod logs:
https://github.com/project-codeflare/codeflare-operator/blob/main/.github/workflows/e2e_tests.yaml#L95-L99

src/codeflare_sdk/templates/base-template.yaml
```diff
@@ -16,21 +16,21 @@ spec:
       custompodresources:
       - replicas: 1
         requests:
-          cpu: 2
```
sutaakar, line 19: Are the template changes expected?

sutaakar (commented):
/lgtm

As intermittent solution to have some e2e tests up and running.

tests/e2e/mnist_raycluster_sdk_test.go
```diff
@@ -0,0 +1,242 @@
...
+	job, err := test.Client().Core().BatchV1().Jobs(namespace.Name).Create(test.Ctx(), job, metav1.CreateOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+	test.T().Logf("Created Job %s/%s successfully", job.Namespace, job.Name)
+
+	go func() {
+		// Checking if pod is found and running
+		podName := ""
+		foundPod := false
+		for !foundPod {
+			pods, _ := test.Client().Core().CoreV1().Pods(namespace.Name).List(test.Ctx(), metav1.ListOptions{
```
astefanutti, line 202: I'd rather avoid that go routine with a for loop that's hard to read. As it's for waiting for the Pod to be available, I'd suggest to either:
- use a "raw" Pod instead of a Job to run the SDK
- or use `Eventually` to do the polling until the Pod can be found.

astefanutti:
> > @ChristianZaccaria Have you considered whether it has more sense for e2e test to be implemented in Python instead of Go, to install and leverage SDK directly?
> 
> I think @astefanutti suggested we kept it as a go test, but this could change.
> 
> We were discussing that in this PR we currently have a well working e2e test and that it might be best to merge it to gate PRs and unblock dependabot updates.

Right, this PR has been open for quite some times already, so now that we have a working e2e test, we would want to merge it ASAP. We can rework the test logic in Python in a separate PR, but that's lower priority compared to having the e2e test suite running in CI.

src/codeflare_sdk/templates/base-template.yaml
ChristianZaccaria: Thanks for pointing that out. I can't remember why I had to modify those values for the test, I think it was until another PR needed to get merged. I reverted the values and the test continues to pass now.

jbusche:
I tried this out on my Mac connected to my OpenShift Cluster, and it worked after I made 2 adjustments.
1.  The SED on my MAC doesn't have the same syntax as Linux, so to get the make commands to work I had to edit the Makefile and change:
```
SED ?= /usr/bin/sed
to
SED ?= /opt/homebrew/bin/gsed
```
2. The Ray operator wasn't able to find it's image, so I had to do this:
```
oc edit deploy kuberay-operator -n ray-system
and  change:
 image: kuberay/operator:v1.0.0-rc.1
to
 image: quay.io/kuberay/operator:v1.0.0-rc.1
```
At that point, then the ray operator image can start and the e2e test can run and succeed.
```
go test -timeout 30m -v ./tests/e2e
# github.com/project-codeflare/codeflare-sdk/tests/e2e.test
ld: warning: -bind_at_load is deprecated on macOS
ld: warning: '/private/var/folders/_4/270ms5_50qq81ql6yt7f27d00000gn/T/go-link-1664163050/go.o' has malformed LC_DYSYMTAB, expected 122 undefined symbols to start at index 147609, found 142 undefined symbols starting at index 83
=== RUN   TestMNISTRayClusterSDK
=== PAUSE TestMNISTRayClusterSDK
=== CONT  TestMNISTRayClusterSDK
    mnist_raycluster_sdk_test.go:49: Created ConfigMap test-ns-dncvt/config-gfp5d successfully
    mnist_raycluster_sdk_test.go:84: Created ServiceAccount test-ns-dncvt/sa-8zmvh successfully
    mnist_raycluster_sdk_test.go:85: Created Role test-ns-dncvt/role-phv6m successfully
    mnist_raycluster_sdk_test.go:86: Created RoleBinding test-ns-dncvt/role-phv6m successfully
    mnist_raycluster_sdk_test.go:195: Created Job test-ns-dncvt/sdk successfully
    mnist_raycluster_sdk_test.go:232: Waiting for Job test-ns-dncvt/sdk to complete
    mnist_raycluster_sdk_test.go:214: Waiting for pod to start...
    mnist_raycluster_sdk_test.go:209: Pod is running!
    test.go:134: Retrieving Pod Container test-ns-dncvt/mnist-head-wvw4m/ray-head logs
    test.go:122: Creating ephemeral output directory as CODEFLARE_TEST_OUTPUT_DIR env variable is unset
    test.go:125: Output directory has been created at: /var/folders/_4/270ms5_50qq81ql6yt7f27d00000gn/T/TestMNISTRayClusterSDK1609076889/001
    test.go:134: Retrieving Pod Container test-ns-dncvt/mnist-worker-small-group-mnist-27b27/machine-learning logs
    test.go:134: Retrieving Pod Container test-ns-dncvt/sdk-jlrd2/test logs
--- PASS: TestMNISTRayClusterSDK (296.83s)
PASS
ok  	github.com/project-codeflare/codeflare-sdk/tests/e2e	297.377s
```

jbusche (approved):
LGTM

ChristianZaccaria:
> I tried this out on my Mac connected to my OpenShift Cluster, and it worked after I made 2 adjustments.
> 
> 1. The SED on my MAC doesn't have the same syntax as Linux, so to get the make commands to work I had to edit the Makefile and change:
> 
> ```
> SED ?= /usr/bin/sed
> to
> SED ?= /opt/homebrew/bin/gsed
> ```
Thanks a lot for testing it out and for pointing those details out! I see the `SED` command comes from CodeFlare-Operator's makefile, perhaps we could make it check for the OS and assign the SED path accordingly. 



> 
> 2. The Ray operator wasn't able to find it's image, so I had to do this:
> 
> ```
> oc edit deploy kuberay-operator -n ray-system
> and  change:
>  image: kuberay/operator:v1.0.0-rc.1
> to
>  image: quay.io/kuberay/operator:v1.0.0-rc.1
> ```
> 
> At that point, then the ray operator image can start and the e2e test can run and succeed.
> 
> ```
> --- PASS: TestMNISTRayClusterSDK (296.83s)
> PASS
> ok  	github.com/project-codeflare/codeflare-sdk/tests/e2e	297.377s
> ```

In terms of the KubeRay image I'm not really sure. When running this in KinD, the pod uses [this image](https://hub.docker.com/r/kuberay/operator/tags?page=1&name=v1.0.0-rc.1) when running `make setup-e2e`:
```
image: docker.io/kuberay/operator:v1.0.0-rc.1
```
Thanks again James!

PR #313, opened 2023-08-16
Release workflow opens PR with changes

### LINKED ISSUE #290: Release workflow should open a PR with changes instead of direct push
Right now the release workflow pushes the release changes directly into the branch.
This approach doesn't work out of the box due to the branch protection in repository. As a result branch protection has to be disabled during release.

Release workflow should open a PR instead of pushing changes directly and wait until the PR is merged by OpenShift CI.

Also release workflow should release to Pypi repository as a last step, to make sure that the release flow will pass. The reason is that Pypi release cannot be deleted/reverted.

Also make sure that all created release artefacts (i.e. created release on GitHub and such) are removed in case release workflow fails.

jiripetrlik:
# Issue link

Closes #290 

# What changes have been made

- Release workflow opens PR with changes instead of direct push
- All publish actions were moved to the end of workflow

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,6 +29,8 @@ jobs:
         runs-on: ubuntu-latest
         permissions:
             contents: write
+        env:
+            PR_BRANCH_NAME: adjustments-release-${{ github.event.inputs.release-version }}
         steps:
             - name: Checkout the repository
               uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


sutaakar (approved):

sutaakar:
@jiripetrlik please check the failing Pr check, there is some formatting issue

anishasthana:
I opened #315 which has the same changes as this PR. It also includes Jiri's specific commit (but with the formatting fix) -- we should close this PR out.

jiripetrlik:
Closing this PR because it was replaced by https://github.com/project-codeflare/codeflare-sdk/pull/315/commits

PR #318, opened 2023-08-21
Add is_dashboard_ready function + update unit test

### LINKED ISSUE #314: Update wait_ready() behavior (or add a separate check for dashboard accessibility)
## WHY
Currently, there is a gap between when the RayCluster itself is ready (and `cluster.wait_ready()` completes), and when the dashboard is accessible (happens a couple seconds later at most). This means that CI or scripts where job submission is performed immediately after `wait_ready()` has a small chance to fail, depending on how long the dashboard takes to be ready.
## WHAT
Either wait_ready has to wait for the dashboard to be ready, or we can add a second function that waits for dashboard availability. Anything that can be used in CI/scripts to block until job submission is possible.

ChristianZaccaria:
# Issue link

Closes #314 

# What changes have been made

- Added an `is_dashboard_ready` function that is used to check for the dashboard's availability. This is done by sending a request to the Dashboard URI, where if status code equals 200 we return `True`.
- This new function is called within the `wait_ready` function to ensure that both the cluster and the dashboard are operational.

# Verification steps

I updated the unit tests to reflect these changes on the `wait_ready` function. All tests are passing.

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -38,6 +38,7 @@ from .model import (
 from kubernetes import client, config
 import yaml
 import os
+import requests
 
 
 class Cluster:
@@ -250,6 +251,13 @@ class Cluster:
 
         return status, ready
 
+    def is_dashboard_ready(self) -> bool:
+        response = requests.get(self.cluster_dashboard_uri(), timeout=5)
+        if response.status_code == 200:
+            return True
+        else:
+            return False
+
     def wait_ready(self, timeout: Optional[int] = None):
         """
         Waits for requested cluster to be ready, up to an optional timeout (s).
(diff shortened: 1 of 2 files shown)


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -217,27 +218,38 @@ def status(
...
                     "WARNING: Current cluster status is unknown, have you run cluster.up yet?"
                 )
-            if not ready:
+            if not ready or not dashboard_ready:
                 if timeout and time >= timeout:
                     raise TimeoutError(f"wait() timed out after waiting {timeout}s")
                 sleep(5)
                 time += 5
-        print("Requested cluster up and running!")
+        print("Requested cluster dashboard are up and running!")
```
Bobbins228, line 252:
```suggestion
        print("Requested cluster and dashboard are up and running!")
```
Small nit pick

Bobbins228 (changes requested):
Tested changes and everything works as expected just two small nit picks with the cluster ready message

src/codeflare_sdk/cluster/cluster.py
ChristianZaccaria: Nice catch, thanks Mark! That's committed now.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -217,27 +218,38 @@ def status(
...
         return status, ready
 
+    def is_dashboard_ready(self) -> bool:
+        try:
+            response = requests.get(self.cluster_dashboard_uri(), timeout=5)
+            if response.status_code == 200:
+                return True
+        except requests.RequestException:
+            pass
+        return False
```
MichaelClifford, line 228: Does this need to be a try/except? requests.get() should always return some status_code, right? 
Can we change it to be a bit simpler:

```python
response = requests.get(self.cluster_dashboard_uri(), timeout=5)
if response.status_code == 200:
        return True
else:    
        return False
```

MichaelClifford (changes requested):

Maxusmusti:
This will also need a rebase

MichaelClifford (approved):
lgtm

PR #327, opened 2023-08-28
[WIP] set head node ip in jobs

MichaelClifford:
# Issue link
https://github.com/project-codeflare/torchx/issues/4 

# What changes have been made
In working through the upstreaming of our changes, I discovered that we could move some of the logic into the SDK itself. This is better for our use case as we have specific needs and the ability to control job submissions that the average torchx users  likely does not have. 

 I've added a new function in `Cluster`, `get_head_ip` that returns the ip address of the head ray node within the cluster. 

I've added a new function to `DDPJobDefinition`, `_set_rdzv_as_head_node` that updates the torchx cmd arguments to force the use of the head node as the rdzv endpoint. 

I also rearranged how `_dry_run` calls ddp to allow editing of its outputs prior to a job submission. 

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -313,6 +313,22 @@ class Cluster:
                 return f"http://{route['spec']['host']}"
         return "Dashboard route not available yet, have you run cluster.up()?"
 
+    def get_head_ip(self) -> str:
+        """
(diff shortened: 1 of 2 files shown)


KPostOffice:
ddp has been removed from the SDK. Please reopen if this PR is still applicable

PR #332, opened 2023-08-29
remove scheduling spec

MichaelClifford:
# Issue link

# What changes have been made

Found bug where "schedulingSpec" is added to the appwrapper without the use of a propriety class. This can cause the Ray Clusters to get stuck in a creation/deletion loop. 

Updated the base template and changed how the schedulingSpec get created to ensure its only added to the appwrapper along with priorityClasses. 

# Verification steps

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests


Maxusmusti (approved):
LGTM

PR #337, opened 2023-09-01
Rework SDK priority implementation + updated testing

Maxusmusti:
# Issue link
#96 

# What changes have been made
Reworked priority class SDK implementation, and now also update the priority value. Also updated tests.

# Verification steps
Create Ray Clusters successfully both with and without priorities.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/tests/unit_test_support.py
+++ b/tests/unit_test_support.py
@@ -46,7 +46,7 @@ def createClusterConfig():
         instascale=True,
         machine_types=["cpu.small", "gpu.large"],
         image_pull_secrets=["unit-test-pull-secret"],
-        dispatch_priority="default",
+        # dispatch_priority="default",
     )
     return config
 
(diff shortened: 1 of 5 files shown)


tests/unit_test_support.py
```diff
@@ -46,7 +46,7 @@ def createClusterConfig():
         instascale=True,
         machine_types=["cpu.small", "gpu.large"],
         image_pull_secrets=["unit-test-pull-secret"],
-        dispatch_priority="default",
+        # dispatch_priority="default",
```
MichaelClifford, line 49: do you want to just remove this line?

Maxusmusti: oh good catch, can do

Maxusmusti: ok should be fixed now

MichaelClifford (approved):
LGTM

PR #356, opened 2023-09-25
Check for dashboard readiness after cluster is ready

### LINKED ISSUE #355: is_dashboard_ready error causes wait_ready to crash
### Describe the Bug

Based on this [Slack thread](https://project-codeflare.slack.com/archives/C04LK0G441F/p1695412369130759?thread_ts=1695412285.357469&cid=C04LK0G441F) by @Maxusmusti 

The `is_dashboard_ready` check errors out when the cluster/appwrapper is still in a pending state, causing `wait_ready` to crash.

### Steps to Reproduce the Bug

1. Run `cluster.up()` and `cluster.wait_ready()` together.
2. See the error.

### Potential solution
Checking if the dashboard is ready AFTER the cluster is ready, instead of both at every timestep from the start.

ChristianZaccaria:
# Issue link

Resolves #355 

# What changes have been made

The check for `is_dashboard_ready` is now performed after the cluster is ready.

# Verification steps

1. Run `cluster.up()` and `cluster.wait_ready()` together.
2. See no errors.

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -274,19 +274,27 @@ class Cluster:
         dashboard_ready = False
         status = None
         time = 0
-        while not ready or not dashboard_ready:
+        while not ready:
             status, ready = self.status(print_to_console=False)
-            dashboard_ready = self.is_dashboard_ready()
             if status == CodeFlareClusterStatus.UNKNOWN:
                 print(
                     "WARNING: Current cluster status is unknown, have you run cluster.up yet?"
                 )
-            if not ready or not dashboard_ready:
+            if not ready:
                 if timeout and time >= timeout:
-                    raise TimeoutError(f"wait() timed out after waiting {timeout}s")
+                    raise TimeoutError(f"wait() timed out after waiting {timeout}s for cluster to be ready")
                 sleep(5)
                 time += 5
-        print("Requested cluster and dashboard are up and running!")
+        print("Requested cluster is up and running!")
+
+        while not dashboard_ready:
+            dashboard_ready = self.is_dashboard_ready()
+            if not dashboard_ready:
+                if timeout and time >= timeout:
+                    raise TimeoutError(f"wait() timed out after waiting {timeout}s for dashboard to be ready")
+                sleep(5)
+                time += 5
+        print("Dashboard is ready!")
 
     def details(self, print_to_console: bool = True) -> RayCluster:
         cluster = _copy_to_ray(self)
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -264,7 +264,7 @@ class Cluster:
         else:
             return False
 
-    def wait_ready(self, timeout: Optional[int] = None):
+    def wait_ready(self, timeout: Optional[int] = None, dashboard_check: bool = True):
         """
         Waits for requested cluster to be ready, up to an optional timeout (s).
         Checks every five seconds.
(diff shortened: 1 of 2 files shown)


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -268,19 +268,27 @@ def wait_ready(self, timeout: Optional[int] = None):
...
                 print(
                     "WARNING: Current cluster status is unknown, have you run cluster.up yet?"
                 )
-            if not ready or not dashboard_ready:
+            if not ready:
+                if timeout and time >= timeout:
+                    raise TimeoutError(f"wait() timed out after waiting {timeout}s for cluster to be ready")
+                sleep(5)
+                time += 5
+        print("Requested cluster is up and running!")
```
ChristianZaccaria, line 282: Open to thoughts on whether I should create a helper function to avoid repetition. A similar pattern is used when checking for dashboard readiness.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -268,19 +268,27 @@ def wait_ready(self, timeout: Optional[int] = None):
...
+
+        while not dashboard_ready:
```
Maxusmusti, line 284: While we are separating the two, could we also add a new argument for wait_ready like `dashboard_check: bool=True` (default true) and have this line be like:
```
while dashboard_check and not dashboard_ready:
```
Just for kubernetes folks who may not have/want a dashboard route currently

Maxusmusti (changes requested):
Looks good, just one request that I know may be worth adding for the future

Maxusmusti:
Also, looks like pre-commit test failed, may need to run. `black` on changed files (or let the pre-commit module make the change locally)

src/codeflare_sdk/cluster/cluster.py
ChristianZaccaria: Great idea, that's added there now. One thought on that though, the `while dashboard_check and not dashboard_ready` is checking if the dashboard is ready if the `dashboard_check` is true. But, whether it's true or false, the route will be generated and can be seen when running i.e., `cluster.details()`. Is this something to assess in this PR or in a separate one?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -268,19 +268,27 @@ def wait_ready(self, timeout: Optional[int] = None):
...
+            dashboard_ready = self.is_dashboard_ready()
+            if not dashboard_ready:
                 if timeout and time >= timeout:
-                    raise TimeoutError(f"wait() timed out after waiting {timeout}s")
+                    raise TimeoutError(f"wait() timed out after waiting {timeout}s for dashboard to be ready")
                 sleep(5)
                 time += 5
-        print("Requested cluster and dashboard are up and running!")
+        print("Dashboard is ready!")
```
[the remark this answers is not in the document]
ChristianZaccaria: Nice catch, thanks! I used `if dashboard_ready` instead as it seemed more appropriate because it will never be true if `dashboard_check` is false. What do you think?

src/codeflare_sdk/cluster/cluster.py
Maxusmusti: Yeah this is fine as-is, once Mark's PR on the KubeRay side is complete we will revisit the route itself from the SDK

src/codeflare_sdk/cluster/cluster.py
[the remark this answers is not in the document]
Maxusmusti: Yup, that works!

Maxusmusti (approved):
LGTM

PR #358, opened 2023-09-25
WIP: Adding direct KubeRay compatibility to the SDK

Maxusmusti:
# Issue link
Need to create an issue still

# What changes have been made
Working on adding direct KubeRay support (generate and manage RayClusters)

# Verification steps
Manual test of regular flow, as well as new `mcad=False` in `ClusterConfiguration` flow

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -199,38 +261,39 @@ def status(
...
+                    if appwrapper.status == AppWrapperStatus.PENDING:
+                        status = CodeFlareClusterStatus.QUEUED
+                    else:
+                        status = CodeFlareClusterStatus.QUEUEING
+                    if print_to_console:
+                        pretty_print.print_app_wrappers_status([appwrapper])
+                    return (
+                        status,
+                        ready,
+                    )  # no need to check the ray status since still in queue
```
KPostOffice, line 296: These questions are outside the scope of the changes, but:
Why is `ready` being set in all of these blocks if it is already `False`?
Why not simplify the return statements to just `return False, CodeFlareClusterStatus.Foo`
Why does the print only happen in the last conditional?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -276,7 +339,7 @@ def wait_ready(self, timeout: Optional[int] = None, dashboard_check: bool = True
         time = 0
         while not ready:
             status, ready = self.status(print_to_console=False)
-            if status == CodeFlareClusterStatus.UNKNOWN:
+            if self.config.mcad and status == CodeFlareClusterStatus.UNKNOWN:
```
KPostOffice, line 342: Inside the conditional on L300 you could remove the check for `cluster.status == RayClusterStatus.UNKNOWN` and set `status` equal to starting that way this logging can still be used in the case where the Ray cluster doesn't exist at all.

KPostOffice (commented):
A few comments/questions

src/codeflare_sdk/cluster/cluster.py
Maxusmusti: These issues date all the way back to the original demo version of the SDK, and just haven't been touched since. It would be good to have a separate PR cleaning up the cluster.status() logic a bit

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -598,6 +662,7 @@ def _map_to_ray_cluster(rc) -> Optional[RayCluster]:
 
     config_check()
     api_instance = client.CustomObjectsApi(api_config_handler())
+    # UPDATE THIS
```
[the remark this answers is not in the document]
Maxusmusti: No, just a reminder that this is another point where we are using an OpenShift specific api that may not be obvious

MichaelClifford:
Tested! works as expected. Thanks @Maxusmusti

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -472,7 +549,7 @@ def get_current_namespace():  # pragma: no cover
             return None
 
 
-def get_cluster(cluster_name: str, namespace: str = "default"):
+def get_cluster(cluster_name: str, namespace: str = "default", mcad=True):
```
MichaelClifford, line 552: I think this needs to check for an appwrapper with the same cluster name to decide whether or not the param mcad=True. Otherwise, you could set up a raycluster only cluster earlier, then re-connect here without specifying mcad=False and generate an appwrapper yaml that was never used.

MichaelClifford (dismissed):

src/codeflare_sdk/cluster/cluster.py
Maxusmusti: The issue here is that the AppWrapper CRD may not (and most likely will not) even exist for users who are doing KubeRay direct. Can you test what currently happens if you specify the opposite version (mcad true when you did false, mcad false when you did true), and does it have any unexpected behavior?

MichaelClifford: I think this is what I mean. A user who is using KubeRay direct will not have an associated appwrapper. In the case where there is no appwrapper, set `mcad=false`. I tested the case where I followed the KubeRay direct route, and then re-connected with `get_cluster(mcad=true)` and it overwrote my yaml file with an Appwrapper when it was deployed as just a RayCluster.yaml

Maxusmusti: So the issue is that the user may not have the codeflare operator or MCAD installed at all, so in that case it won't be about a specific AW not existing, its the notion of an AW (the CRD itself with the associated API) won't exist. I guess we should just check for that or try to handle that case as well for auto-selection?

Bobbins228 (changes requested):
Really cool stuff I was able to successfully create ray clusters with and without mcad!
Just needs a rebase now and I can approve :+1:

Bobbins228 (approved):
Ran through after rebase all is as it should nice work /approve

Maxusmusti:
Thanks @Bobbins228, just rebased on Kevin's dashboard oauth changes, if everything still works as expected then it should be good to go!

Issue #365, opened 2023-10-02
Stop SDK from saving Appwrapper yaml to local dir


dimakis:

MichaelClifford:
@dimakis what is the alternative here? saving them to a `/clusters/` dir or something?

dimakis:
@MichaelClifford hiding the appwrapper by default, but allowing for it to be printed if requested.

VanillaSpoon:
As the generated `.yaml` is intended for immediate processing within the same notebook environment, saving directly to the remote file system may remain more efficient. This avoids the download-upload cycle and reduces the steps involved in the workflow.

After a discussion we have decided to still achieve an abstraction, without resorting to local storage, we could move the appwrappers to a `.codeflare/appwrapper/<aw.yaml>` directory within the current notebook. Removing the direct view to the appwrapper but keeping a similar structure.

PR #380, opened 2023-10-11
add: abstraction to remove appwrapper visiblilty from NB console. 

VanillaSpoon:
# Issue link
closes https://github.com/project-codeflare/codeflare-sdk/issues/365

# What changes have been made
To further abstract the appwrapper from a users perspective, they appwrapper generation has been moved to a hidden folder `~/.codeflare/appwrapper/<appwrapper.yaml>.`

# Verification steps
`poetry build` 
Install the new .whl for the sdk in a jupyter notebook. 
Open one of the quick-start demos: `git clone https://github.com/project-codeflare/codeflare-sdk`

Follow the process as usual. 

After your clusterconfiguration ensure this is logged:
`Written to: /opt/app-root/src/.codeflare/appwrapper/<appwrapper.yaml>`

The yaml will no longer appear in the listed files.

From the notebook terminal, you can `cd` to `~/.codeflare/appwrapper/` and then `ls`. This will list the generated .yaml. Feel free to inspect it with `vi <appwrapper-name.yaml>` and ensure the appwrapper name and information is correct. 

Continue with  `cluster.up()` and ensure the appwrapper is applied and raycluster is deployed as expected, with the correct cluster info in the kuberay-operator pod logs.

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -71,8 +71,10 @@ class Cluster:
         """
         self.config = config
         self.app_wrapper_yaml = self.create_app_wrapper()
-        self.app_wrapper_name = self.app_wrapper_yaml.split(".")[0]
         self._job_submission_client = None
+        self.app_wrapper_name = self.app_wrapper_yaml.replace(".yaml", "").split("/")[
+            -1
+        ]
 
     @property
     def _client_headers(self):
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -19,6 +19,7 @@ This sub-module exists primarily to be used internally by the Cluster object
 
 import yaml
 import sys
+import os
 import argparse
 import uuid
 from kubernetes import client, config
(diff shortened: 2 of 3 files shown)


Fiona-Waters:
Ran through verification steps and can confirm that appwrapper yaml file is written to .codeflare directory. Ran a demo notebook and all works as expected.
/lgtm

Maxusmusti (changes requested):
One thing to note, I think this should probably be in an absolute path, `~/.codeflare/...` (or expand ~ accordingly to user home dir), rather than a relative path based on current working directory. This way we match the kubernetes pattern of `~/.kube/...`

Bobbins228 (commented):
Tested this and it works as expected
/lgtm

KPostOffice:
Sorry for the late review, but if we're trying to abstract the creation of the appwrapper from the user, what is the purpose of creating a yaml file at all? Why not just generate the raw text into memory at cluster creation time?

Maxusmusti:
@KPostOffice Adding the slack comment here as well:

It's something @MichaelClifford, @dimakis, and I have previously discussed, essentially providing the actual yamls themselves is still useful for a lot of users that may want to tweak them with options the SDK doesn't currently support, check them before submitting to their cluster, re-use them, etc.

At the same time, using relative paths was a bit messy, and there were users (mainly BU), who didn't want the AppWrappers generated in their current working dir. So to both organize all codeflare-related assets, as well as to clean/hide implementation details for those who don't care while keeping them accessible for those who do, we ended up deciding on copying the kubernetes ~/.kube/ pattern with ~/.codeflare/ dir for asset management and accessibility

VanillaSpoon:
Thanks for raising the discussion on this @KPostOffice, and thank you @Maxusmusti for providing additional information :)

KPostOffice (approved):
/lgtm

PR #407, opened 2023-11-17
Fixed creating ingresses without admin access

Bobbins228:
# Issue link

Closes #400 
# What changes have been made

Notable changes:
1. Removed the use of getting the ingress domain from `config.openshift.io.ingresses` and opted for the creation of routes when on OpenShift. 
2.  Fixed the creation of ingresses when setting `mcad=False` in cluster configuration
3. On OpenShift users must pass an ingress_domain when using the ray client. (this is mandatory as we need the domain name when creating the cert command in the yaml see [here](https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/utils/generate_yaml.py#L443)
4. We now use the discovery api to discern whether the user is on Kind or OpenShift by searching for the `route.openshift.io` api group. (This should resolve cluster level privileges issue)

# Verification steps

1. Run through a notebook on Kind & OpenShift and verify ingresses/routes are created properly
2. Run through a notebook on Kind with `mcad=False` to verify ingresses are created/deleted
## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #432, opened 2023-12-13
Fixed get_cluster() method

### LINKED ISSUE #431: Failing to get Clusters info using get_cluster() function for the Ray cluster created with Ingress configuration
### Describe the Bug
Failing to get Cluster details using get_cluster() function for the Ray cluster created with Ingress configuration in kind clusters thorwing eror as **ingress_domain is invalid**
A clear and concise description of the bug.

**Error:**

```
Traceback (most recent call last):
  File "mnist_rayjob.py", line 12, in <module>
    cluster = get_cluster('mnist',namespace)
  File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 648, in get_cluster
    return Cluster.from_k8_cluster_object(rc, mcad=mcad)
  File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 513, in from_k8_cluster_object
    return Cluster(cluster_config)
  File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 73, in __init__
  File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/cluster/cluster.py", line 195, in create_app_wrapper
    return generate_appwrapper(
  File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/utils/generate_yaml.py", line 667, in generate_appwrapper
    update_dashboard_ingress(
  File "/opt/app-root/lib64/python3.8/site-packages/codeflare_sdk/utils/generate_yaml.py", line 142, in update_dashboard_ingress
    raise ValueError(
ValueError: ingress_domain is invalid. For Kubernetes Clusters please specify an ingress domain
```

### Codeflare Stack Component Versions
Please specify the component versions in which you have encountered this bug.

Codeflare SDK:v0.12.1

### What Have You Already Tried to Debug the Issue?

On Kind clusters ingress_domain is a mandatory field but there is no ingress_domain option added in the [from_k8s_cluster_object](https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/cluster/cluster.py#L471) function within get_cluster()

### Expected Behavior

get_cluster() function should return the ray cluster details successfully for clusters with Ingress domain created

Bobbins228:
# Issue link

Closes #431 
# What changes have been made

Ingresses have the label `ingress-owner: appwrapper` which is used to identify them. 
When `mcad=True` the SDK gets the the ownership reference from the ingress object to identify and check it is a match with the appwrapper name. 

When `mcad=False` the ingresses are listed and the domain is gathered from the ingress with the `ingress-owner` label. 
For both scenarios the ingress domain is extracted from the host of the ingress.

Included steps to recreate scenarios where ingress_options are set.

Fixed missing ingress creation steps for the `_component_resources_up` and `_component_resources_down` functions for cases when `mcad=False`.
# Verification steps

Run through a demo notebook and create a Ray Cluster on Kind and OpenShift.
run `cluster = get_cluster(name, namespace)` 
`cluster.down()` and other `cluster.` commands should work.

Create a Ray Cluster on Kind with ingress options set see [here](https://github.com/project-codeflare/codeflare-sdk?tab=readme-ov-file#codeflare-sdk-in-a-kubernetes-environment).
Run `cluster = get_cluster(name, namespace)` 

Create a Ray cluster with `mcad=False` an ingress should be created now and `get_cluster()` should work too.
## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -492,7 +492,7 @@ class Cluster:
             to_return["requirements"] = requirements
         return to_return
 
-    def from_k8_cluster_object(rc, mcad=True):
+    def from_k8_cluster_object(rc, mcad=True, ingress_domain=None):
         machine_types = (
             rc["metadata"]["labels"]["orderedinstance"].split("_")
             if "orderedinstance" in rc["metadata"]["labels"]
@@ -532,6 +532,7 @@ class Cluster:
             ]["image"],
             local_interactive=local_interactive,
             mcad=mcad,
+            ingress_domain=ingress_domain,
         )
         return Cluster(cluster_config)
 
@@ -685,7 +686,29 @@ def get_cluster(cluster_name: str, namespace: str = "default"):
     for rc in rcs["items"]:
         if rc["metadata"]["name"] == cluster_name:
             mcad = _check_aw_exists(cluster_name, namespace)
-            return Cluster.from_k8_cluster_object(rc, mcad=mcad)
+
+            try:
+                config_check()
+                api_instance = client.NetworkingV1Api(api_config_handler())
+                ingresses = api_instance.list_namespaced_ingress(namespace)
+                if mcad == True:
+                    for ingress in ingresses.items:
+                        # Search for ingress with AppWrapper name as the owner
+                        if cluster_name == ingress.metadata.owner_references[0].name:
+                            ingress_host = ingress.spec.rules[0].host
+                else:
+                    for ingress in ingresses.items:
+                        # Search for the ingress with the ingress-owner label
+                        if ingress.metadata.labels["ingress-owner"] == cluster_name:
+                            ingress_host = ingress.spec.rules[0].host
+            except Exception as e:
+                return _kube_api_error_handling(e)
+
+            # We gather the ingress domain from the host
+            ingress_domain = ingress_host.split(".", 1)[1]
+            return Cluster.from_k8_cluster_object(
+                rc, mcad=mcad, ingress_domain=ingress_domain
+            )
     raise FileNotFoundError(
         f"Cluster {cluster_name} is not found in {namespace} namespace"
     )
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -691,6 +691,7 @@ def get_cluster(cluster_name: str, namespace: str = "default"):
                 config_check()
                 api_instance = client.NetworkingV1Api(api_config_handler())
                 ingresses = api_instance.list_namespaced_ingress(namespace)
+                ingress_host = None
                 if mcad == True:
                     for ingress in ingresses.items:
                         # Search for ingress with AppWrapper name as the owner
@@ -705,7 +706,11 @@ def get_cluster(cluster_name: str, namespace: str = "default"):
                 return _kube_api_error_handling(e)
 
             # We gather the ingress domain from the host
-            ingress_domain = ingress_host.split(".", 1)[1]
+            if ingress_host is not None:
+                ingress_domain = ingress_host.split(".", 1)[1]
+            else:
+                ingress_domain = None
+
             return Cluster.from_k8_cluster_object(
                 rc, mcad=mcad, ingress_domain=ingress_domain
             )
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -492,7 +492,7 @@ class Cluster:
             to_return["requirements"] = requirements
         return to_return
 
-    def from_k8_cluster_object(rc, mcad=True, ingress_domain=None):
+    def from_k8_cluster_object(rc, mcad=True, ingress_domain=None, ingress_options={}):
         machine_types = (
             rc["metadata"]["labels"]["orderedinstance"].split("_")
             if "orderedinstance" in rc["metadata"]["labels"]
@@ -502,6 +502,10 @@ class Cluster:
             "volumeMounts"
             in rc["spec"]["workerGroupSpecs"][0]["template"]["spec"]["containers"][0]
         )
+        if local_interactive:
+            ingress_domain = get_ingress_domain_from_client(
+                rc["metadata"]["name"], rc["metadata"]["namespace"]
+            )
         cluster_config = ClusterConfiguration(
             name=rc["metadata"]["name"],
             namespace=rc["metadata"]["namespace"],
@@ -533,6 +537,7 @@ class Cluster:
             local_interactive=local_interactive,
             mcad=mcad,
             ingress_domain=ingress_domain,
+            ingress_options=ingress_options,
         )
         return Cluster(cluster_config)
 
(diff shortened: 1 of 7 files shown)


sutaakar:
> The get_cluster() function now asks for ingress_domain for the case of Kind clusters. It defaults at None and is not required on OpenShift clusters.

Why does `get_cluster()` function need to know Ingress domain? The cluster already exists, Ingress was applied when cluster was created.

Bobbins228:
@sutaakar 
>Why does get_cluster() function need to know Ingress domain? The cluster already exists, Ingress was applied when cluster was created.

The `get_cluster()` function uses the `from_k8_cluster_object(rc, mcad=True)` function to create a new [Cluster Configuration](https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/cluster/cluster.py#L481-L510)

We use the Ray cluster spec `rc` to gather info like image, name, machine_types, etc.
Unfortunately we cannot gather the domain name which was used originally to create the cluster because it is not included in the Ray cluster spec and was put in place when the ingress object was created in the AppWrapper.

This is necessary as we plan to reintroduce routes in #407 and we will not be supplying domains for them as OpenShift does so automatically.

Alternatively we could go under the assumption that the ingress is created successfully and use `list_namespaced_object` to gather the ingress and extract the domain from the host that way. If that doesn't work we can ask the user to supply the ingress name in `get_cluster` WDYT?

sutaakar:
@Bobbins228 I am still not sure what do you need the domain for when calling `get_cluster()`.
If you create a cluster then you need this value to create Ingress resource. But for `get_cluster()`, the Ingress already exists, you don't need to create it any more. So, why is the value needed?

Do you need domain to compute what is the Dashboard URL, or is there any other reason?

Bobbins228:
@sutaakar 
When `from_k8_cluster_object` creates that new ClusterConfiguration an appwrapper is created the exact same way we would create one before calling `cluster.up()` see ClusterConfiguration init [here](https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/cluster/cluster.py#L73)

To create this app wrapper we need all required fields to be fulfilled in the case of Kind clusters `ingress_domain` is required despite the fact the cluster and ingresses have already been created previously.

`from_k8_cluster_object` does most of the job filling in the appwrapper for us but `ingress_domain` is not as simple to fill in automatically.

sutaakar:
@Bobbins228 What will happen in case user fills some random value as `ingress_domain`, i.e. "abcde.com"?
From your description it looks like the value is solely needed to fulfill internal implementation requirements without any functionality overlaps.

Bobbins228:
@sutaakar 
I see your point in the case the user provides the wrong `ingress_domain` I think you could still use it for any `cluster.` functions but ultimately the aw is wrong in that scenario.

It is not ideal hence the suggestion to first check for the ingress in the namespace and gather the domain from the host before asking for the `ingress_domain`.

KPostOffice:
Yeah, I'm not a huge fan of this either. Could you add an annotation to the ray-cluster that has the ingress domain and then get it directly from there?

sutaakar:
Another option could be to assign an owner for Ingress/Route (if it doesn't have) and use it to query for Ingress belonging to the AppWrapper or Ray cluster.

Bobbins228:
@KPostOffice @sutaakar 
I have updated the PR,

We now get the appwrapper object and extract the domain from the ingress Host when `mcad=True` and in the opposite case we get the ingress object that has the `ingress-owner: cluster_name` label and extract the domain from the host there.

You no longer have to provide the `ingress_domain` for `get_cluster`

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -644,7 +645,40 @@ def get_cluster(cluster_name: str, namespace: str = "default"):
...
+                except Exception as e:  # pragma: no cover
+                    return _kube_api_error_handling(e, print_error=False)
+            else:
+                # Case where mcad is false we get the host from the created ingress.
+                try:
+                    config_check()
+                    api_instance = client.NetworkingV1Api(api_config_handler())
+                    ingresses = api_instance.list_namespaced_ingress(namespace)
+                    for ingress in ingresses.items:
+                        if ingress.metadata.labels["ingress-owner"] == cluster_name:
```
sutaakar, line 672: Have you consider using [Owner functionality](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) instead of custom label?

Bobbins228:
@sutaakar I was able to apply the owner functionality in relation to ray clusters created with app wrappers but ingresses created when `mcad=False` do not have an owner so I have kept in the label for that case.

sutaakar:
When does the `mcad=False` case happen?

Bobbins228:
@sutaakar 
Yes it is a parameter of the cluster configuration.
By setting `mcad=False` an appwrapper is not created instead a yaml file featuring the RC, Ingress and secret. 
The SDK will use KubeRay to create the ray cluster and individually create the ingress and secrets which would make them have no owner in the traditional sense.

sutaakar:
hmm, so SDK creates RayCluster CR directly. Would it have sense to add owner for Secret and Ingress pointing to RayCluster CR?

Bobbins228:
@sutaakar To add a owner reference I need to add in the uid of the ray cluster right?

sutaakar:
@Bobbins228 Right, uid is required.

Bobbins228:
@sutaakar We won't be able to apply the uid of the Ray Cluster to the ingress because it is created when the rc is up. Ingresses and secrets are created at the same time.

For the case of `mcad=False` we will have to keep the `ingress-owner` label

sutaakar:
@Srihari1192 do you want to review the PR too?

Srihari1192:
> un `cluster = get_cluster(name, namespace)`

Sure will take a look

Srihari1192:
> @Srihari1192 do you want to review the PR too?

Sure will take a look

Srihari1192:
@Bobbins228 @sutaakar I have noticed below error when testing with this changes in this [PR](https://github.com/project-codeflare/codeflare-operator/pull/424/files) at [step](https://github.com/project-codeflare/codeflare-operator/pull/424/files#diff-2f1df199e4bbdee5d99116602dac73f4330603cbda71556ae6d1b33c553d181fR12) get_cluster('mnist',namespace)


`Traceback (most recent call last):
  File "mnist_rayjob.py", line 12, in <module>
    cluster = get_cluster('mnist',namespace)
  File "/codeflare-sdk/codeflare-sdk/src/codeflare_sdk/cluster/cluster.py", line 710, in get_cluster
    ingress_domain = ingress_host.split(".", 1)[1]
IndexError: list index out of range
`

Bobbins228:
@Srihari1192 
Have you set your ingress up with ingress_domain or ingress_options?

Srihari1192:
@Bobbins228 Yes the Ingress_options is set same as in sdk test [here](https://github.com/project-codeflare/codeflare-operator/pull/424/files#diff-1031535ceb5f0388a7c0f2806f2b424d4d28f6ebade12641003801701662e695R27-R41) and also i tried verifying existing [sdk tests](https://github.com/project-codeflare/codeflare-operator/pull/424/files#diff-c3831e4439ca2188655e58e2f468b2a2b9d29766b0b2f047eca5fad914170265R56-R58) same error here aswell https://github.com/project-codeflare/codeflare-operator/actions/runs/7406823001/job/20151892482?pr=424

Bobbins228:
@Srihari1192 What is put in place of the host for `CLUSTER_HOSTNAME`?
https://github.com/project-codeflare/codeflare-operator/pull/424/files#diff-1031535ceb5f0388a7c0f2806f2b424d4d28f6ebade12641003801701662e695R11

Srihari1192:
> @Srihari1192 What is put in place of the host for `CLUSTER_HOSTNAME`? https://github.com/project-codeflare/codeflare-operator/pull/424/files#diff-1031535ceb5f0388a7c0f2806f2b424d4d28f6ebade12641003801701662e695R11

It will be set based on the cluster type here https://github.com/project-codeflare/codeflare-operator/pull/424/files#diff-572dd9104509ce659f6f38075e4ad54935cb91373ab4a03c3a05eaa91f4c39c8R190-R206

Bobbins228:
@Srihari1192 
Okay there is an issue here with the way we get the ingress domain from the host.
[Here](https://github.com/project-codeflare/codeflare-sdk/pull/432/files#diff-5706a96c5346dd54a22370211260411363c4ed41d798fddf240eb8983b1366eaR708-R712) we get the host from the ingress associated with the Ray Cluster and try to get the domain by extracting the string after a `.` 

I will have a look further to see if there is a better way to extract the ingress domain from the host.

KPostOffice (approved):

PR #447, opened 2024-01-19
Added directory creation check for cases where mcad=False

Bobbins228:
# Issue link

# What changes have been made

Added a conditional statement that will create the `~/.codeflare/appwrapper` directory if it doesn't exist for the `write_components()` function.
# Verification steps

Ensure the `~/.codeflare/appwrapper` directory does not exist
Run poetry build on the PR changes and install the whl file
Run through an SDK demo notebook but set `mcad=False` in the cluster configuration.
The folder and yaml file should be present at `~/.codeflare/appwrapper`
## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -619,6 +619,11 @@ def _create_oauth_sidecar_object(
 
 
 def write_components(user_yaml: dict, output_file_name: str):
+    # Create the directory if it doesn't exist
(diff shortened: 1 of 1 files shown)


astefanutti (approved):
Can we also make it so the directory is renamed to something different from `appwrapper` in a follow up PR?

Fiona-Waters (approved):
Ran through verification steps and directory is created as expected.
/approve

PR #449, opened 2024-01-23
Add RayCLuster SDK Oauth Authentication test

Srihari1192:
# Issue link

closes https://issues.redhat.com/browse/RHOAIENG-55
# What changes have been made

- Added tests for RayCluster SDK Oauth Authentication scenarios with and without login using python
- Added pytest marker for each tests for selecting tests to run on kind or openshift cluster environment
- Refactored setup methods kube configuration , create and delete namespace in support.py 

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


tests/e2e/mnist_raycluster_sdk_auth_test.go
```diff
@@ -0,0 +1,439 @@
...
+	}
+
+	// Create cluster wide RBAC, required for SDK OpenShift Oauth check
+	clusterPolicyRules := []rbacv1.PolicyRule{
+		{
+			Verbs:         []string{"get", "list"},
+			APIGroups:     []string{"config.openshift.io"},
+			Resources:     []string{"ingresses"},
+			ResourceNames: []string{"cluster"},
+		},
```
sutaakar, line 90: Is this Role still needed?

tests/e2e/mnist_raycluster_sdk_auth_test.go
```diff
@@ -0,0 +1,439 @@
...
+	job, err := test.Client().Core().BatchV1().Jobs(namespace.Name).Create(test.Ctx(), job, metav1.CreateOptions{})
+	test.Expect(err).NotTo(HaveOccurred())
+	test.T().Logf("Created Job %s/%s successfully", job.Namespace, job.Name)
+
+	// Setup the codeflare-sdk inside the pod associated to the created job
+	SetupCodeflareSDKInsidePod(test, namespace, job.Name)
+
+	test.T().Logf("Checking pod logs contains insecure requests warning")
+	// Assert the job pod as warning of insecurerequests
+	assertPodLogs(test, namespace, job.Name, "InsecureRequestWarning: Unverified HTTPS request is being made to host", 3*time.Minute)
```
sutaakar, line 215: I would expect SDK to provide some information about the issue, not just `InsecureRequestWarning` which is not relevant to the actual problem.

sutaakar: Thinking whether we should explicitly check for the dashboard/API availability. This line verifies that SDK returns a warning, possibly as part of `cluster.wait_ready()` call. However implementation of this call can change making this test incorrect.

Would it have sense to add specific REST call in Python script invoking Dashboard/Jobs API, making sure those call returns correct response code for unauthenticated user?

tests/e2e/mnist_raycluster_sdk_auth_test.go
Srihari1192: Its not needed, will remove it , Thanks

tests/e2e/mnist_raycluster_sdk_auth_test.go
Srihari1192: Yes I think its make sense to have check to correct unauthorised response code and message

Srihari1192: @sutaakar @Bobbins228 @KPostOffice  I tried to Submit Ray job in multiple ways for testing unauthenticated scenario but getting same below response error as in UI as login should required to connect with Ray dashboard. 

```
resp=requests.post(dashboard_url+"/api/jobs/",
                   json={ "entrypoint":"python mnist.py",
    "runtime_env":{"working_dir": "workdir","pip": "requirements.txt"}})

client = RayJobClient(dashboard_url)
submission_id = client.submit_job(
    entrypoint="python mnist.py",
    runtime_env={"working_dir": "./","pip": "mnist_pip_requirements.txt"},
)

 An unexpected error occurred: HTTPSConnectionPool(host='ray-dashboard-test.com', port=443): Max retries exceeded with url: /api/jobs/ (Caused by SSLError(SSLCertVerificationError(1, [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1007)')))
```

I think for Authentication tests we can go with current implementation approach with cluster token and verify Ray Job submission with authentication and  for  non authentication tests as we don't get exact error saying that Not authorised to submit Job, either we can have negative check incase of submitting job to ray dashboard which is not allowing to submit by validating response code other than 200 or skipping this test scenario .

sutaakar: @Srihari1192 Have you tried to disable certificate verification?
https://stackoverflow.com/questions/15445981/how-do-i-disable-the-security-certificate-check-in-python-requests

Srihari1192: @sutaakar By disabling certificate verification  able to test this will update the PR

Srihari1192:
Will convert this test to python tests similar to PR https://github.com/project-codeflare/codeflare-sdk/pull/451

Bobbins228 (commented):
I ran through these changes with `poetry run pytest -v -s ./tests/e2e/mnist_raycluster_sdk_oauth_test.py -m openshift` and I received this error.

```
tests/e2e/mnist_raycluster_sdk_oauth_test.py:65: in run_mnist_raycluster_sdk_oauth
    self.assert_jobsubmit_withlogin(cluster)
tests/e2e/mnist_raycluster_sdk_oauth_test.py:105: in assert_jobsubmit_withlogin
    status = job.status()
../.local/lib/python3.11/site-packages/codeflare_sdk/job/jobs.py:201: in status
    return self._runner.status(self._app_handle)
../.local/lib/python3.11/site-packages/torchx/runner/api.py:419: in status
    desc = scheduler.describe(app_id)
../.local/lib/python3.11/site-packages/torchx/schedulers/ray_scheduler.py:383: in describe
    job_status_info = self._get_job_status(app_id)
../.local/lib/python3.11/site-packages/torchx/schedulers/ray_scheduler.py:376: in _get_job_status
    client = self._get_ray_client(job_submission_netloc=addr)
 ```
With the final pytest error being:
```FAILED tests/e2e/mnist_raycluster_sdk_oauth_test.py::TestRayClusterSDKOauth::test_mnist_ray_cluster_sdk_auth - ValueError: client netloc (ray-dashboard-mnist-test-ns-6t2pj.apps.rosa.mcampbel.n399.p3.openshiftapps.com) does not match job netloc (ray)```

Did you come across something like this when testing @Srihari1192

**Comment by Srihari1192** [comment]
> poetry run pytest -v -s ./tests/e2e/mnist_raycluster_sdk_oauth_test.py

no it works fine on my cluster 
I didnt added [import ray](https://github.com/project-codeflare/codeflare-sdk/blob/361a9ec461f1394b7b2fcb32e41b689b265e0749/tests/e2e/mnist_raycluster_sdk_test.py#L8) as assumed not required , is it failing because of this

File: tests/e2e/mnist_raycluster_sdk_oauth_test.py [hunk 3]
```diff
@@ -0,0 +1,158 @@
...
+            )
+        )
+
+        cluster.up()
+        self.assert_appwrapper_exists()
+
+        cluster.status()
+
+        cluster.wait_ready()
+        self.assert_raycluster_exists()
```
Bobbins228, line 57: Is calling this function redundant? 
If `cluster.wait_ready()` passes it is safe to assume the Ray Cluster was created.

Bobbins228 (commented):
Re-ran `poetry install --with test,docs` 
Test passed just one nit

tests/e2e/mnist_raycluster_sdk_oauth_test.py
Srihari1192: I noticed we were doing similar check in [sdk test ](https://github.com/Srihari1192/codeflare-sdk/blob/361a9ec461f1394b7b2fcb32e41b689b265e0749/tests/e2e/mnist_raycluster_sdk_test.py#L90), so i kept same here aswell

Srihari1192: In assertion we are calling ray API call to check resource is created or not , not sure we are performing same check  in cluster.wait_ready()

Bobbins228: I suppose its okay to keep its just that `wait_ready` would fail if a ray cluster didn't exist. :+1:

Bobbins228 (commented):
/lgtm

tests/e2e/mnist_raycluster_sdk_oauth_test.py
Srihari1192: @Bobbins228  yes it makes sense to skip this check, updated the code

Bobbins228 (commented):
/lgtm

tests/e2e/mnist_raycluster_sdk_test.py
```diff
@@ -14,30 +14,24 @@
...
-        if hasattr(self, "namespace"):
-            self.api_instance.delete_namespace(self.namespace)
+        delete_namespace(self)
 
     def test_mnist_ray_cluster_sdk(self):
-        self.create_test_namespace()
+        create_namespace(self)
         self.run_mnist_raycluster_sdk()
 
     def create_test_namespace(self):
```
ChristianZaccaria, line 37: I believe this whole function can be removed as it is now in the `support.py` file.

ChristianZaccaria (changes requested):
Just one nitpick there. After that, lgtm

ChristianZaccaria (changes requested):
It seems to just need a rebase, afterwards, /lgtm

sutaakar:
@Srihari1192 Can you adjust https://github.com/project-codeflare/codeflare-sdk/blob/main/.github/workflows/e2e_tests.yaml#L126 to run all tests with proper `kind` marker?

sutaakar:
Can you add timeout for tests?
From my local experiments it seems that https://pypi.org/project/pytest-timeout/ works quite well.

sutaakar (approved):
/lgtm

Bobbins228 (approved):
Awesome stuff Srihari works like a charm :1st_place_medal: 
/approve

PR #451, opened 2024-01-30
Convert Go e2e test to Python Test to leverage SDK directly

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-2058 

# What changes have been made

- Converted Go e2e test to a Python test to leverage the SDK directly.
- Removed unnecessary RBAC, ConfigMap, and Job specs.
- Removed Python script and moved its contents directly into the test file, this way allowing to perform assertions in the test at any given point.
- Adjusted `ingress_options` in the script to include this annotation: `nginx.ingress.kubernetes.io/proxy-body-size: "100M"`, else, on KinD, we would encounter the following error:
```
E       RuntimeError: Request failed with status code 413: <html>
E       <head><title>413 Request Entity Too Large</title></head>
```
- Adjusted e2e workflow + added `sdk-user` with limited permissions on KinD cluster to run the e2e tests.

# Verification steps

- This PR: https://github.com/project-codeflare/codeflare-common/pull/30 should be merged first to test the e2e workflow as the KinD cluster is created based on that configuration.
- This PR: https://github.com/project-codeflare/codeflare-operator/pull/456 should be merged first for testing the e2e test locally (or add changes in codeflare-operator `kind.sh` script).

To run the **e2e test locally on KinD cluster**, follow these steps:

Note: If not already done so, please add in your local `/etc/hosts` file `127.0.0.1 kind`. This will map your localhost IP address to the KinD cluster's hostname. This is already performed on [GitHub Actions](https://github.com/project-codeflare/codeflare-common/blob/1edd775e2d4088a5a0bfddafb06ff3a773231c08/github-actions/kind/action.yml#L70-L72)

- Setup Phase:
  - Pull the [codeflare-operator repo](https://github.com/project-codeflare/codeflare-operator) and run the following make targets:
  ```
  make kind-e2e
  export CLUSTER_HOSTNAME=kind
  export CODEFLARE_TEST_TIMEOUT_LONG=20m
  make deploy -e IMG=quay.io/project-codeflare/codeflare-operator:v1.1.0
  make setup-e2e
  ```
- Test Phase:
   - Once we have the codeflare-operator and kuberay-operator running and ready, we can run the e2e test on the codeflare-sdk repo:
  ```
  poetry install --with test,docs
  poetry run pytest -v -s ./tests/e2e/mnist_raycluster_sdk_test.py
  ```

To run the **e2e test locally on OpenShift**, follow these steps:
- Setup Phase:
  - Pull the [codeflare-operator repo](https://github.com/project-codeflare/codeflare-operator) and run the following make targets:
  ```
  make deploy -e IMG=quay.io/project-codeflare/codeflare-operator:v1.1.0
  make setup-e2e
  ```
- Test Phase:
   - Once we have the codeflare-operator and kuberay-operator running and ready, we can run the e2e test on the codeflare-sdk repo:
  ```
  poetry install --with test,docs
  poetry run pytest -v -s ./tests/e2e/mnist_raycluster_sdk_test.py
  ```

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


tests/e2e/mnist_raycluster_sdk_test.py
```diff
@@ -0,0 +1,152 @@
...
+        self.run_mnist_raycluster_sdk()
+
+    def create_test_namespace(self):
+        self.namespace = f"test-ns-{random_choice()}"
+        namespace_body = client.V1Namespace(metadata=client.V1ObjectMeta(name=self.namespace))
+        self.api_instance.create_namespace(namespace_body)
+        return self.namespace
+
+    def run_mnist_raycluster_sdk(self):
+        ray_image = "quay.io/project-codeflare/ray:latest-py39-cu118"
```
sutaakar, line 49: Can you allow providing custom image using env variable?

tests/e2e/support.py
```diff
@@ -0,0 +1,14 @@
+import random
+import string
+
+def read_file(file_name):
```
sutaakar, line 4: Is this function used anywhere?

tests/e2e/mnist_raycluster_sdk_test.py
ChristianZaccaria: I added that now, thanks!

Bobbins228 (commented):
/lgtm

tests/e2e/mnist_raycluster_sdk_test.py
```diff
@@ -0,0 +1,171 @@
...
+            self.api_instance.api_client
+        )
+
+    def teardown_method(self):
+        if hasattr(self, "namespace"):
+            self.api_instance.delete_namespace(self.namespace)
+        if hasattr(self, "configmap"):
+            self.api_instance.delete_namespaced_config_map(
+                self.configmap.metadata.name, self.namespace
+            )
```
sutaakar, line 41: What ConfigMap is deleted here? I didn't find any other reference for `configmap` attribute.

ChristianZaccaria: Good catch, forgot to remove it, we no longer need to create one - Removed

sutaakar (commented):
/lgtm

VanillaSpoon:
This is great :) Ran the tests locally also and they run through as expected
\lgtm

astefanutti (commented):
What user is impersonated to run the tests? It seems it's using the default user that's cluster admin, in which case that's a regression from the existing e2e tests?

sutaakar:
https://github.com/project-codeflare/codeflare-common/pull/33 is adding support for custom user. That can be used to specify user with specific allowed roles.
For usage see https://github.com/project-codeflare/codeflare-operator/pull/462

ChristianZaccaria:
@astefanutti @sutaakar , I was told to keep the issues separated where after merging we would add the unprivileged user to run the test: https://issues.redhat.com/browse/RHOAIENG-60 

We could add it in the scope of this PR too if preferred.

astefanutti:
> I added in the e2e workflow an `sdk-user` with limited RBAC permissions who runs the e2e tests. This is currently only performed in the e2e test workflow. For local testing, should I add in the test a script that creates the `sdk-user` + permissions if it doesn't already exist in the cluster, or should we use the cluster admin for local tests? Thanks @sutaakar @astefanutti

For local testing, I'd avoid mixing provisioning and testing logic, so an option would be to document how to setup the cluster (even specific to KinD would be fine) and run the tests with the correct user.

sutaakar (commented):
/lgtm

astefanutti:
/approve

Great job!

Bobbins228 (approved):
/approve

PR #455, opened 2024-02-09
add: simplified import statements for the codeflare-sdk

VanillaSpoon:
# Issue link
closes https://issues.redhat.com/browse/RHOAIENG-2109

# What changes have been made
This pr introduces a simplified import approach for the codeflare-sdk. Reducing the nested element of import statements. 
For example: 
```py
    from codeflare_sdk.cluster.cluster import Cluster, ClusterConfiguration
    from codeflare_sdk.cluster.auth import TokenAuthentication
```

Becomes: 
```py
    from codeflare_sdk import Cluster, ClusterConfiguration, TokenAuthentication
```

This pr also contains the relevent updates to documentation and example notebooks.

# Verification steps
Build the sdk and install it on your notebook.
Run through the examples, with the updated import statements, and ensure everything runs.

## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #458, opened 2024-02-12
update: Demo notebooks to use simplified import statements

VanillaSpoon:
# Issue link
Closes https://issues.redhat.com/browse/RHOAIENG-2933

# What changes have been made
This is not to be merged until after another stack release, to ensure usability. 

The pr contains updates to the import statements in the documentation, and demo notebooks. Making use of the simplified, non-nested statements from https://github.com/project-codeflare/codeflare-sdk/pull/455

# Verification steps
Run through the demo notebooks and ensure all the imports are functioning correctly.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/demo-notebooks/additional-demos/hf_interactive.ipynb
+++ b/demo-notebooks/additional-demos/hf_interactive.ipynb
@@ -42,8 +42,7 @@
    "outputs": [],
    "source": [
     "# Import pieces from codeflare-sdk\n",
-    "from codeflare_sdk.cluster.cluster import Cluster, ClusterConfiguration\n",
-    "from codeflare_sdk.cluster.auth import TokenAuthentication"
+    "from codeflare_sdk import Cluster, ClusterConfiguration, TokenAuthentication"
    ]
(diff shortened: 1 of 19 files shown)


ChristianZaccaria (changes requested):
For future reminder/tracking, let's make this change on all scripts used by tests including e2e tests. Great work!

ChristianZaccaria:
> Thanks @ChristianZaccaria :) For the updated imports I have specifically targeted user facing functions, this excludes the utils and helper functions that are being used in the tests.
> 
> Optionally I can add all the functions to the `__init__.py`, and this can also allow imports for the tests to be used similarily, or only expose the used functions which will require the test imports to remain as they are.

That makes sense thanks Eoin! I think it would be a good idea as a follow-on to allow it for test items, but also not completely necessary as it's not user facing. /lgtm

Bobbins228 (approved):
/approve

PR #479, opened 2024-03-06
use lookup table for status

KPostOffice:
# Issue link

# What changes have been made
Use lookup table for status

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


Bobbins228 (commented):
Ran through looks good.
Just wondering should we show the `Cluster Queue status` in `cluster.status()` if it is a Ray cluster that is running?

PR #483, opened 2024-03-15
Set Cluster Configuration to operate in memory by default

Bobbins228:
# Issue link

[RHOAIENG-4525](https://issues.redhat.com/browse/RHOAIENG-4525)
# What changes have been made

By default the SDK now handles appwrapper/individual-resource creation in memory.
Yaml files are not generated while `write_to_file=False` in cluster configuration

# Verification steps

* Checkout this PR's branch
## Testing
This is very tedious to test as there are 4 possible ways of creating a cluster config with this change. i.e. (`mcad=False`, `write_to_file=False`), (`mcad=True`, `write_to_file=False`), ...

I recommend to start by testing `mcad=True/False` with `write_to_file=False`
* Delete your `~/.codeflare` folder - Not necessary but easier to catch things working incorrectly.
* Follow a demo notebook of your choosing
* Run `ClusterConfiguration` and a yaml/appwrapper file will not be written you should receive a message like this `Yaml resources loaded for <cluster-name>`
* Continue to run through the demo notebook
Repeat with an alternate mix of the variables.
## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/config.py
+++ b/src/codeflare_sdk/cluster/config.py
@@ -55,3 +55,4 @@ class ClusterConfiguration:
     openshift_oauth: bool = False  # NOTE: to use the user must have permission to create a RoleBinding for system:auth-delegator
     ingress_options: dict = field(default_factory=dict)
     ingress_domain: str = None
+    write_to_file: bool = False
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -73,9 +73,7 @@ class Cluster:
         self.config = config
         self.app_wrapper_yaml = self.create_app_wrapper()
         self._job_submission_client = None
-        self.app_wrapper_name = self.app_wrapper_yaml.replace(".yaml", "").split("/")[
-            -1
-        ]
+        self.app_wrapper_name = self.config.name
 
     @property
     def _client_headers(self):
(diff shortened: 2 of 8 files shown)


VanillaSpoon (commented):
Ran through a demo notebook testing each of the expected outputs as well as default, everything is going smoothly :) code changes look good too.

/lgtm

One thing though, should an explanation be added within docs/cluster-configuration.md?

Bobbins228:
@VanillaSpoon 
Thanks Eoin!

> One thing though, should an explanation be added within docs/cluster-configuration.md?
I plan to make an addition to the cluster config docs after a new release :+1:

src/codeflare_sdk/cluster/config.py
```diff
@@ -55,3 +55,4 @@ class ClusterConfiguration:
     openshift_oauth: bool = False  # NOTE: to use the user must have permission to create a RoleBinding for system:auth-delegator
     ingress_options: dict = field(default_factory=dict)
     ingress_domain: str = None
+    write_to_file: bool = False
```
KPostOffice, line 58: Can we make this an arg to the cluster.up() command instead since it isn't configuration for the actual cluster?

KPostOffice (commented):
Looks good. One small comment

src/codeflare_sdk/cluster/config.py
Bobbins228: The appwrapper creation logic is all handled when a user runs `cluster = Cluster(ClusterConfiguration())` so the `write_to_file` bool needs to be a config param.
https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/cluster/cluster.py#L74

Fiona-Waters:
Successfully ran through all 4 scenarios. All works as expected. Love this addition, great work Mark!
/approve

PR #484, opened 2024-03-18
Upgrade from AppWrapper v1beta1 to v1beta2

dgrove-oss:
# Issue link

# What changes have been made

Context: This PR supports the migration of the codeflare operator from MCADv1 to Kueue+AppWrapper. We should probably wait on merging this PR until the corresponding PRs on the consuming codeflare repos are ready, as the removal of MCADv1 CRDs is a breaking change.

Depends on: https://github.com/project-codeflare/codeflare-common/pull/37

# Verification steps

I have verified that unit tests are running locally.

I expect end-to-end tests will fail because there is a circular dependency between updates to the SDK and updates to the codeflare operator (the e2e tests for this repo install the codeflare operator, which can't be updated until the SDK is updated).

## Checks
- Testing Strategy
   - [ x ] Unit tests
   - [ ] Manual tests


dgrove-oss:
This PR is now ready for review.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -600,9 +551,7 @@ def list_all_queued(namespace: str, print_to_console: bool = True):
     Returns (and prints by default) a list of all currently queued-up AppWrappers
     in a given namespace.
     """
-    app_wrappers = _get_app_wrappers(
-        namespace, filter=[AppWrapperStatus.RUNNING, AppWrapperStatus.PENDING]
-    )
+    app_wrappers = _get_app_wrappers(namespace, filter=[AppWrapperStatus.SUSPENDED])
```
KPostOffice, line 554: Shouldn't this still have `RUNNING` or do the new and old `RUNNING` statuses have different meanings?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -925,16 +874,16 @@ def _get_app_wrappers(
...
         app_wrapper = _map_to_app_wrapper(item)
-        if filter and app_wrapper.status in filter:
-            list_of_app_wrappers.append(app_wrapper)
+        if filter:
+            if app_wrapper.status in filter:
+                list_of_app_wrappers.append(app_wrapper)
         else:
-            # Unsure what the purpose of the filter is
             list_of_app_wrappers.append(app_wrapper)
     return list_of_app_wrappers
```
KPostOffice, line 882: This can be simplified to:
```
if not filter or app_wrapper.status in filter:
    list_of_app_wrappers.append(app_wrapper)
```

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -1010,18 +959,14 @@ def _map_to_ray_cluster(rc) -> Optional[RayCluster]:
...
-            can_run=aw["status"]["canrun"],
-            job_state=aw["status"]["queuejobstate"],
+            status=AppWrapperStatus(aw["status"].get("phase", "suspended").lower()),
         )
     return AppWrapper(
         name=aw["metadata"]["name"],
-        status=AppWrapperStatus("queueing"),
-        can_run=False,
-        job_state="Still adding to queue",
+        status=AppWrapperStatus("suspended"),
```
KPostOffice, line 969: Can you just use AppWrapperStatus.SUSPENDED directly here in order to avoid magic string?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -1010,18 +959,14 @@ def _map_to_ray_cluster(rc) -> Optional[RayCluster]:
...
 
 def _map_to_app_wrapper(aw) -> AppWrapper:
-    if "status" in aw and "canrun" in aw["status"]:
+    if "status" in aw:
         return AppWrapper(
             name=aw["metadata"]["name"],
-            status=AppWrapperStatus(aw["status"]["state"].lower()),
-            can_run=aw["status"]["canrun"],
-            job_state=aw["status"]["queuejobstate"],
+            status=AppWrapperStatus(aw["status"].get("phase", "suspended").lower()),
```
KPostOffice, line 965: `"suspended"` -> `AppWrapperStatus.SUSPENDED.value`

tests/test-case-bad.yaml
```diff
@@ -1,167 +1,148 @@
-apiVersion: workload.codeflare.dev/v1beta1
+apiVersion: workload.codeflare.dev/v1beta2
 kind: AppsWrapper
 metadata:
   labels:
     orderedinstance: cpu.small_gpu.large
   nam: unit-test-cluster
   namspace: ns
 spec:
-  priority: 9
```
KPostOffice, line 9: Quick Q. Is priority removed from AppWrapper completely?

KPostOffice (commented):
Thanks @dgrove-oss . A few comments

.github/workflows/e2e_tests.yaml
```diff
@@ -64,11 +64,12 @@ jobs:
         with:
           repository: project-codeflare/codeflare-operator
           path: codeflare-operator
+          ref: mcadv1b2
```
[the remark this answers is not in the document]
dgrove-oss: This is only because the CFO PR isn't merged.  Once we merge all the pieces, this goes back to pulling from main.

src/codeflare_sdk/cluster/cluster.py
dgrove-oss: I wasn't sure what the desired semantics really were here.  The function is called `list_all_queued` and the comment says
```
    """
    Returns (and prints by default) a list of all currently queued-up AppWrappers
    in a given namespace.
    """
```
Neither old or new AppWrapper should be including `Running` in the filter for this, because Running means that the AppWrapper has been dispatched (it is not in the queue) and it is executing.  If we really did intend to include Running, then we should change the comment (and probably the name of the function).

src/codeflare_sdk/cluster/cluster.py
dgrove-oss: Done.

.github/workflows/e2e_tests.yaml
```diff
@@ -64,11 +64,12 @@ jobs:
...
 
       - name: Set Go
         uses: actions/setup-go@v5
         with:
-          go-version: v1.20
+          go-version: v1.21
```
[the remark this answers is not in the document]
dgrove-oss: Good catch.  Not sure why I missed doing that here.

src/codeflare_sdk/cluster/cluster.py
dgrove-oss: Done.

tests/test-case-bad.yaml
dgrove-oss: Yes.  We dropped all AppWrapper-specific priority support.   One would now just use Kueue's workload priority classs mechanisms (https://kueue.sigs.k8s.io/docs/concepts/workload_priority_class/) for AppWrappers just like you'd do with any other resource type that Kueue can manage.

dgrove-oss: We probably should consider adding support for Kueue's workload priorities to the SDK, but it should be done fresh since it works differently and it would apply to everything (not just AppWrappers).

src/codeflare_sdk/cluster/cluster.py
KPostOffice: ++, cant argue with you on that one especially since it's not used internally at all. I think your implementation is more clear w.r.t. implied functionality.

dgrove-oss:
Closing as outdated.

PR #480, opened 2024-03-12
Remove creation of OAuth resources/logic and remove openshift_oauth option

ChristianZaccaria:
# Issue link

Issue: https://issues.redhat.com/browse/RHOAIENG-1993 

# What changes have been made

- This PR removes the logic in the SDK that creates the OAuth resources.
- Remove `openshift_oauth` from ClusterConfiguration as defaulting to True.
- Added `COOKIE_SECRET` env for oauth proxy container. - [Here](https://github.com/project-codeflare/codeflare-operator/pull/453#:~:text=%2D%20%2D%2Dcookie%2Dsecret%3D%24(COOKIE_SECRET))
- Removed unit tests related to the creation/deletion of OAuth resources.
- Adjusted test cases and unit tests.
- Verified this change works with the RayCluster Controller
- Add `verify_tls` option to ClusterConfiguration: this option is set to True by default but still has the option to change for debugging. Based on this option the endpoint will be verified, if set to false, it alters if the user verifies that the certificate is trusted.

# Verification steps

To test these changes:
0. In an OpenShift cluster.
1. Checkout this PR [RayCluster Controller](https://github.com/project-codeflare/codeflare-operator/pull/453). In the CFO repository you can run the CFO in your cluster with `make run NAMESPACE=default` and `make install`.
2. Deploy and install KubeRay
3. Checkout this PR and install the SDK
4. Run through [this notebook](https://github.com/project-codeflare/codeflare-sdk/blob/main/demo-notebooks/guided-demos/2_basic_jobs.ipynb), change between verify_tls=True/False.
5. Once `cluster.up()` is ran, the oauth-proxy sidecar is created by default if on an OpenShift cluster. You should find several resources created including a ServiceAccount named `jobtest-oauth-proxy` and a route named `ray-dashboard-jobtest` which were created by the RayCluster Controller as opposed by the SDK.
6. The route should be accessible through authentication.

## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #494, opened 2024-04-02
Converted Go upgrade tests in Codeflare-SDK to python tests and add support for kueue integration

abhijeet-dhumal:
# Issue link

[RHOAIENG-2444](https://issues.redhat.com/browse/RHOAIENG-2444)

# What changes have been made

- [X] - Convert Go upgrade tests in Codeflare-SDK to python tests.
- [X] - Remove Go-related files since they are no longer required
- [X] - Update upgrade test in accordance with Kueue integration.
- [X] - Add support method to create kueue resources.
- [X] - Add PR check step to apply kueue CRD's and add cluster-roles to get/add/delete kueue resources.

# Verification steps

- Install Kueue in cluster
`oc apply --server-side -k "github.com/opendatahub-io/kueue/config/rhoai?ref=dev"`
- Create a 'Cluster-queue' resource (LocalQueue will be created in test-namespace as a part of test execution)
- Test RayClusterSDKUpgradeTest in Openshift cluster (_Execute in same order_)
`poetry run pytest -v -s ./tests/upgrade/raycluster_sdk_upgrade_test.py::TestMNISTRayClusterUp`
 ~ _asserts RayCluster status - `READY` (Namespace created in `TestMNISTRayClusterUp` will be used in `TestMnistJobSubmit` )_
`poetry run pytest -v -s ./tests/upgrade/raycluster_sdk_upgrade_test.py::TestMnistJobSubmit` 
~ _asserts JobSubmission status_

**Note** :  In case of testing with Kind cluster, skip [token authentication step](https://github.com/project-codeflare/codeflare-sdk/pull/494/files#diff-43b7b3e97d52443b407b45657443b44ceb967d93c62d0e7344917fde406bc076R32-R37).

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests

**- Manual testing followed :** 
-
image
-
image
image
-
image
-


ChristianZaccaria (changes requested):
Great work Abhijeet! Since we no longer require the use of `.go` files we can remove `go.mod`, `go.sum`, and the `support.go` file too.

Srihari1192 (commented):
@abhijeet-dhumal Both the tests `TestMNISTRayClusterUp` and `TestMnistJobSubmit` should run independently for upgrade testing , incorporate accordingly.  
 you can run tests independently like this `poetry run pytest -v -s ./tests/e2e/raycluster_sdk_upgrade_test.py::TestMnistJobSubmit`

abhijeet-dhumal:
> Great work Abhijeet! Since we no longer require the use of `.go` files we can remove `go.mod`, `go.sum`, and the `support.go` file too.

@ChristianZaccaria Thanks for reviewing, yes I have cleaned up Go files since they are no longer required

abhijeet-dhumal:
> @abhijeet-dhumal Both the tests `TestMNISTRayClusterUp` and `TestMnistJobSubmit` should run independently for upgrade testing , incorporate accordingly. you can run tests independently like this `poetry run pytest -v -s ./tests/e2e/raycluster_sdk_upgrade_test.py::TestMnistJobSubmit`

Thank you for reviewing @Srihari1192 :) 

According to above review I have made some changes wrt independent execution of tests and thoroughly tested required usecases of upgrade tests. 

image

image

tests/e2e/raycluster_sdk_upgrade_test.py
```diff
@@ -0,0 +1,172 @@
...
+import pytest
+
+from support import *
+from codeflare_sdk.cluster.cluster import get_cluster
+
+namespace = "test-ns-rayupgrade"
+
+
+# Creates a Ray cluster
+@pytest.mark.openshift
```
Srihari1192, line 18: remove this marking as this test wont be run as part of e2e

Srihari1192 (changes requested):
Keep this tests in same upgrade folder as if we run the tests with command `poetry run pytest -v -s ./tests/e2e` this tests also be picked and not the right way

tests/e2e/raycluster_sdk_upgrade_test.py
abhijeet-dhumal: I was facing an error while importing support functions in separate upgrade folder as the codeflare-sdk init file wasn't there to be used as a package
```
from tests.e2e.support import *
E   ModuleNotFoundError: No module named 'tests'
```

Adding empty __init__ files in tests and upgrade folder resolved the import error !!
Is this approach seems feasible?

tests/upgrade/raycluster_sdk_upgrade_test.py
```diff
@@ -0,0 +1,171 @@
...
+                head_cpus="1",
+                head_memory=2,
+                min_cpus="1",
+                max_cpus=1,
+                min_memory=1,
+                max_memory=2,
+                num_gpus=0,
+                instascale=False,
+                image=ray_image,
+                openshift_oauth=True,
```
ChristianZaccaria, line 50: `openshift_oauth` can be removed here, I believe it will be removed automatically after rebasing.

We are now defaulting to always use openshift_oauth so this option has been removed.
https://github.com/project-codeflare/codeflare-sdk/pull/480

ChristianZaccaria (changes requested):

tests/upgrade/raycluster_sdk_upgrade_test.py
```diff
@@ -0,0 +1,164 @@
...
+                head_memory=2,
+                min_cpus="1",
+                max_cpus=1,
+                min_memory=1,
+                max_memory=2,
+                num_gpus=0,
+                instascale=False,
+                image=ray_image,
+                write_to_file=True,
+                mcad=False,
```
Bobbins228, line 53: Nit: `mcad=False` is not necessary anymore it is default behaviour :)

tests/e2e/support.py
```diff
@@ -45,3 +53,24 @@ def run_oc_command(args):
...
+        "spec": {"clusterQueue": cluster_queue},
+    }
+    print("LocalQueue doesn't exists, creating LocalQueue!")
+    self.custom_api.create_namespaced_custom_object(
+        group="kueue.x-k8s.io",
+        namespace=self.namespace,
+        plural="localqueues",
+        version="v1beta1",
+        body=json,
+    )
```
Bobbins228, line 76: We could do with an exception here in case it is not possible to create the local queue

tests/e2e/support.py
```diff
@@ -24,6 +24,14 @@ def create_namespace(self):
     self.api_instance.create_namespace(namespace_body)
 
 
+def create_namespace_with_name(self, namespace_name):
+    self.namespace = namespace_name
+    namespace_body = client.V1Namespace(
+        metadata=client.V1ObjectMeta(name=self.namespace)
+    )
+    self.api_instance.create_namespace(namespace_body)
```
Bobbins228, line 32: We could do with an exception here in case it is not possible to create the namespace

Bobbins228 (changes requested):

tests/e2e/support.py
abhijeet-dhumal: Thank you @Bobbins228 for reviewing
Yes, these Nit-pick changes seems valid, I will make chnages accordingly.

tests/e2e/support.py
```diff
@@ -45,3 +59,27 @@ def run_oc_command(args):
...
+        self.custom_api.create_namespaced_custom_object(
+            group="kueue.x-k8s.io",
+            namespace=self.namespace,
+            plural="localqueues",
+            version="v1beta1",
+            body=json,
+        )
+    except Exception as e:
+        print(f"An unexpected error occurred while creating 'local-queue'. Error: ", e)
+        delete_namespace(self)
```
Bobbins228, line 85: Same point as above except call `delete_namespace(self)` before `_kube_api_error_handling` in the exception

Bobbins228 (changes requested):

tests/upgrade/raycluster_sdk_upgrade_test.py
```diff
@@ -0,0 +1,165 @@
...
+            )
+            if response.status_code == 403:
+                assert True
+            else:
+                response.raise_for_status()
+                assert False
+
+        except Exception as e:
+            print(f"An unexpected error occurred. Error: {e}")
+            assert False
```
sutaakar, line 120: AFAIK you can reuse `RayJobClient`  instead of REST call, just don't provide any `Authorization` header

tests/upgrade/raycluster_sdk_upgrade_test.py
```diff
@@ -0,0 +1,165 @@
...
+        logs = client.get_job_logs(submission_id)
+        print(logs)
+
+        self.assert_job_completion(status)
+
+        client.delete_job(submission_id)
+        cluster.down()
+
+    def assert_job_completion(self, status):
+        if status == "SUCCEEDED":
```
sutaakar, line 160: Wouldn't it have more sense to assert directly content of `status`?

abhijeet-dhumal: I synced these changes from [SDK-auth test](https://github.com/project-codeflare/codeflare-sdk/commit/88b2518db3657411635e647301e59dfdd65d75a7#diff-0d12faeac6160d6a2a47811c5f68013d2f459bf90d61151091b291e6782f5200R48). I think there are explicit print statements to have job_completion status for which the condition is added. 
But, Yes right! Can assert status directly 👀
Thank you @sutaakar for reviewing!

sutaakar (approved):

Srihari1192 (approved):

tests/upgrade/raycluster_sdk_upgrade_test.py
```diff
@@ -0,0 +1,154 @@
...
+
+from tests.e2e.support import *
+from codeflare_sdk.cluster.cluster import get_cluster
+
+namespace = "test-ns-rayupgrade"
+
+
+# Creates a Ray cluster
+class TestMNISTRayClusterUp:
+    def setup_method(self):
```
ChristianZaccaria, line 19: Perhaps we could include a teardown as well to delete the namespace after the test is ran, something like:
```
    def teardown_method(self):
        delete_namespace(self)
```

ChristianZaccaria (changes requested):
Great work! Left a few comments

tests/upgrade/raycluster_sdk_upgrade_test.py
sutaakar: Not in this case as the test is intended to verify upgrade.
So the intended use case is to create a cluster, upgrade operator and/or SDK, get cluster and make sure we still can work with it.

abhijeet-dhumal: As a part of upgrade test, we need namespace and raycluster created during `TestRayClusterUp` to be used by `Test_mnist_job_submission`, right? 
If we include this tear-down method in `TestRayClusterUp`, then the created namespace will be deleted and won't be available for `Test_mnist_job_submission`

sutaakar:
@abhijeet-dhumal Can you add roles to create/get?/delete Kueue resources in https://github.com/project-codeflare/codeflare-sdk/blob/a53fc1e617130c3017de50b0ed4681777993eace/.github/workflows/e2e_tests.yaml#L102-L103 ?

.github/workflows/e2e_tests.yaml
```diff
@@ -97,6 +97,9 @@ jobs:
         with:
           user-name: sdk-user
 
+      - name: Add kueue resources
+        run: kubectl apply --server-side -f "https://github.com/kubernetes-sigs/kueue/releases/download/v0.6.2/manifests.yaml"
```
ChughShilpa, line 101: I think we should use the manifests for opendatahub/kueue

ChughShilpa, line 101: I think we should use the manifests from opendatahub-io/kueue

sutaakar: I would consider SDK as upstream, so it should be integrated with upstream components.

sutaakar (approved):
/lgtm

tests/upgrade/raycluster_sdk_upgrade_test.py
```diff
@@ -0,0 +1,154 @@
...
+                namespace=self.namespace,
+                num_workers=1,
+                head_cpus="1",
+                head_memory=2,
+                min_cpus="1",
+                max_cpus=1,
+                min_memory=1,
+                max_memory=2,
+                num_gpus=0,
+                instascale=False,
```
ChughShilpa, line 51: Just a small change, we can remove instascale now

ChughShilpa (changes requested):

tests/upgrade/raycluster_sdk_upgrade_test.py
abhijeet-dhumal: Yes, this update has been made.
Thank you @ChughShilpa for reviewing !

ChughShilpa (approved):
/lgtm

Bobbins228 (commented):
/approve
Ran and verified it works on OpenShift

PR #501, opened 2024-04-04
Fix output logs by switching to kind-cluster context

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-5131 

# What changes have been made

This was previously attempted to be fixed, however when the e2e test fails, the step to `Switch to kind-cluster context` is skipped due to previous step failure.
- With this change, the step to switch context will always be made, enabling to print component logs.

As seen/tested here: https://github.com/project-codeflare/codeflare-sdk/actions/runs/8560772513/job/23460511229?pr=495 

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ChristianZaccaria:
@dgrove-oss awesome, thanks for reviewing!

PR #506, opened 2024-04-10
Add enableIngress and set to false by default

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-5116 

# What changes have been made

- Added `enableIngress` field to `headGroupSpec` and set to false by default.
- Added Forbidden message on attempt to create a RayCluster where the `enableIngress` field is not false.
- The loop iterates over the resources in loaded in yaml file and looks for the RayCluster resource and the enableIngress field. If the enableIngress field is not False, forbidden message is displayed to the user, and if it is False, then we create the resources.

# Verification steps

Should be tested with this Kueue PR - [Can be followed here](https://github.com/opendatahub-io/kueue/pull/26#:~:text=for%20your%20reviewer%3A-,Verification%20Steps%3A,-Login%20to%20your)

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -546,7 +546,7 @@ class Cluster:
     ):
         if self.config.write_to_file:
             with open(self.app_wrapper_yaml) as f:
-                yamls = yaml.load_all(f, Loader=yaml.FullLoader)
+                yamls = list(yaml.load_all(f, Loader=yaml.FullLoader))
                 for resource in yamls:
                     enable_ingress = (
                         resource.get("spec", {})
@@ -559,8 +559,6 @@ class Cluster:
                             f"Forbidden: RayCluster '{name}' has 'enableIngress' set to 'True' or is unset."
                         )
                         return
-                f.seek(0)  # Reset file pointer to the beginning
-                yamls = yaml.load_all(f, Loader=yaml.FullLoader)  # Reload the YAMLs
                 _create_resources(yamls, namespace, api_instance)
         else:
             yamls = yaml.load_all(self.app_wrapper_yaml, Loader=yaml.FullLoader)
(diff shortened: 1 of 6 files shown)


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -547,6 +547,20 @@ def _component_resources_up(
...
+                        .get("enableIngress")
+                    )
+                    if resource["kind"] == "RayCluster" and enable_ingress is not False:
+                        name = resource["metadata"]["name"]
+                        print(
+                            f"Forbidden: RayCluster '{name}' has 'enableIngress' set to 'True' or is unset."
+                        )
+                        return
+                f.seek(0)  # Reset file pointer to the beginning
+                yamls = yaml.load_all(f, Loader=yaml.FullLoader)  # Reload the YAMLs
```
Fiona-Waters, line 563: Why are we loading the yaml here again? I think in the case where this error occurs this function would be run again anyway in order to apply the updated yaml.

Fiona-Waters (commented):
Have run this alongside https://github.com/opendatahub-io/kueue/pull/26 and it works as expected. Just one question.

ChristianZaccaria:
@dimakis RE: can the user modify the `enableIngress` field from an already existing RayCluster resource? 

No, [webhook PR in Kueue](https://github.com/opendatahub-io/kueue/pull/26) also forbids the modification of the `enableIngress` field to anything that is not `false`, including if the user attempts to remove the field entirely would be forbidden.

image

src/codeflare_sdk/cluster/cluster.py
ChristianZaccaria: Previously, the `yaml.load_all` would use a generator which can only be iterated over once, for this reason reloading the YAML was necessary to then be used in the RayCluster creation (probably not the best approach).

Thanks for pointing this out, I now opted to convert it to a list instead in the latest commit, this way we can iterate through it as needed without having to reload the YAML. - Overall behaviour is the same as before.

Some refs: 
https://stackoverflow.com/questions/42802058/pyyaml-load-all-returns-generator-rather-than-a-dict 
https://github.com/yaml/pyyaml/issues/762

ChristianZaccaria:
Rebased - e2e tests passing in this PR

PR #522, opened 2024-04-23
use parse_quantity to convert string to raw resource

KPostOffice:
# Issue link
RHAOIENG 5000

# What changes have been made
Directly use mem string with units rather than attempting to convert back and forth to int. If the actual number of bytes used by a cluster is needed for some reason in the future, use `kubernetes.utils.parse_quantity`

# Verification steps

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -18,6 +18,7 @@ the resources requested by the user. It also contains functions for checking the
 cluster setup queue, a list of all existing clusters, and the user's working namespace.
 """
 
+import re
 from time import sleep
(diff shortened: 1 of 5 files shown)


abhijeet-dhumal (approved):
/lgtm
Manually Ran and verified, using demo notebook.

**Reproducer :** 
In current upstream version, While running 
```
from codeflare_sdk.cluster import get_cluster
print(get_cluster("mnist","default").app_wrapper_name)
```
got error : `ValueError: invalid literal for int() with base 10: '500m'`

This PR resolved above error  and got output as :
 ```
Yaml resources loaded for mnist
mnist
```

PR #526, opened 2024-05-01
ClusterConfiguration parameter that will accept user inputted labels and apply them to the Ray Cluster on creation.

Ygnas:
# Issue link

[RHOAIENG-2017](https://issues.redhat.com/browse/RHOAIENG-2017)

# What changes have been made

Add a ClusterConfiguration parameter that will accept user inputted labels and apply them to the Ray Cluster on creation. 

# Verification steps

Defining the ClusterConfiguration with 3 labels would generate the RayCluster with those labels + any other ones that are already part of it.

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/docs/cluster-configuration.md
+++ b/docs/cluster-configuration.md
@@ -3,7 +3,7 @@
 To create Ray Clusters using the CodeFlare SDK a cluster configuration needs to be created first.<br>
 This is what a typical cluster configuration would look like; Note: The values for CPU and Memory are at the minimum requirements for creating the Ray Cluster.
 
-```
+```python
 from codeflare_sdk import Cluster, ClusterConfiguration
 
 cluster = Cluster(ClusterConfiguration(
@@ -20,8 +20,8 @@ cluster = Cluster(ClusterConfiguration(
     num_gpus=0, # Default 0
     mcad=True, # Default True
     image="quay.io/project-codeflare/ray:latest-py39-cu118", # Mandatory Field
-    instascale=False, # Default False
     machine_types=["m5.xlarge", "g4dn.xlarge"],
+    labels={"exampleLabel": "example", "secondLabel": "example"},
 ))
 ```
 
@@ -30,3 +30,5 @@ From there a user can call `cluster.up()` and `cluster.down()` to create and rem
 
 In cases where `mcad=False` a yaml file will be created with the individual Ray Cluster, Route/Ingress and Secret included.<br>
 The Ray Cluster and service will be created by KubeRay directly and the other components will be individually created.
+
+The `labels={"exampleLabel": "example"}` parameter can be used to apply additional labels to the RayCluster resource.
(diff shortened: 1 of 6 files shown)


ChristianZaccaria (approved):
Changes tested and works as expected:
- Labels added to the generated RayCluster YAML.
- On creation, RayCluster CR contains the supplied labels.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -187,6 +187,7 @@ def create_app_wrapper(self):
         write_to_file = self.config.write_to_file
         verify_tls = self.config.verify_tls
         local_queue = self.config.local_queue
+        user_labels = self.config.user_labels
```
Bobbins228, line 190: Perhaps `cluster_labels` would work better here if just `labels` is not an option?

Bobbins228 (changes requested):

src/codeflare_sdk/cluster/cluster.py
Ygnas: Updated to `cluster_labels` the `labels` are a local var in those functions. Or you think it's still better to call it `labels`?

ChristianZaccaria: That's a good point, I think we should keep it as simple as possible for the user. Just `labels` sounds better to me, and being part of the `ClusterConfiguration` we are already implying it's cluster labels.

Ygnas: Updated to be just `labels`.

Ygnas:
Rebased

docs/cluster-configuration.md
```diff
@@ -22,6 +22,8 @@ cluster = Cluster(ClusterConfiguration(
     image="quay.io/project-codeflare/ray:latest-py39-cu118", # Mandatory Field
     instascale=False, # Default False
     machine_types=["m5.xlarge", "g4dn.xlarge"],
+    # The labels={} parameter can be used to apply additional labels to the RayCluster resource.
```
Fiona-Waters, line 25: I think this should be included in the explanatory paragraph underneath rather than here.

docs/cluster-configuration.md
```diff
@@ -22,6 +22,8 @@ cluster = Cluster(ClusterConfiguration(
     image="quay.io/project-codeflare/ray:latest-py39-cu118", # Mandatory Field
     instascale=False, # Default False
```
Fiona-Waters, line 23: I think this line relating to Instascale should be removed while we are updating this file - please

Fiona-Waters (commented):
Ran a demo notebook on openshift and it works as expected with the labels being added to the generated yaml and the resulting raycluster instance.
 Just a couple of nitpicks but otherwise good to go. Thanks Ignas!

docs/cluster-configuration.md
Ygnas: Removed

Fiona-Waters (commented):
/approve

PR #530, opened 2024-05-08
Updated Ray version to 2.20.0

Bobbins228:
# Issue link

[RHOAIENG-6450](https://issues.redhat.com/browse/RHOAIENG-6450)
# What changes have been made

Updated CFSDK ray dependency to 2.20.0
# Verification steps

* Checkout this PR's branch
## Testing
Use this image for the following test scenarios: `quay.io/project-codeflare/ray:2.20.0-py39-cu118`
* Run through demo notebooks
* Ensure GPU utilization is working correctly
* Ensure basic & local interactive demos work correctly
* Ensure job submission works correctly

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


Fiona-Waters:
Tested these changes through the SDK demo notebooks using quay.io/mcampbel/ray:220-py39-cu118-dev as the ray image.
- Was able to run basic ray demo.
- Was able to run job client demo and use gpus.
- While running basic_interactive demo I see this info message (see screenshot below). I wasn't able to run through the entire demo due to issues with training script which I am currently looking into. But was able to run ray.init successfully. 
- Was able to run local interactive demo - including ray.init but also seeing the info about the python version mismatch
image

This is not an error as such and I was able to submit jobs successfully.

.github/workflows/e2e_tests.yaml
```diff
@@ -43,6 +43,7 @@ jobs:
           sudo rm -rf /opt/ghc
           sudo rm -rf "/usr/local/share/boost"
           sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+          sudo apt install python3.9
```
astefanutti, line 46: Just wondering why it wasn't needed before?

astefanutti (commented):
/lgtm

.github/workflows/e2e_tests.yaml
Bobbins228: Python 3.8 is not supported for Ray versions past 2.10 iirc. 
We were safe before with `ray==2.7.0`.

Bobbins228: Also this brings on another concern in terms of future proofing.

What would we need to do to upgrade the notebook images to python 3.11? 
Seeing as 3.9 is now the oldest supported Python version for the latest Ray packages we don't want to be caught out.

astefanutti: Right, it'd be good to anticipate and loop with the Notebook team.

sutaakar: Python 3.9 installation in e2e tests was added in https://github.com/project-codeflare/codeflare-sdk/pull/542

src/codeflare_sdk/templates/base-template.yaml
```diff
@@ -49,7 +49,7 @@ spec:
           # - kubernetes
         spec:
           # The version of Ray you are using. Make sure all Ray containers are running this version of Ray.
-          rayVersion: '2.7.0'
+          rayVersion: '2.20.0'
```
astefanutti, line 52: Do we want to pin the Ray version in the container image tag as well?

astefanutti: Actually my understanding is that Ray version field is rather pretty loose. It's only used for assembling the command for the Ray job submitter pod: https://github.com/ray-project/kuberay/blob/2c47c2d4d48be52bb6a2693d3aa850a763db5808/ray-operator/controllers/ray/common/job.go#L68.

Now on the point you're raising, I think it should be documented that the client and backend version must be compatible.

demo-notebooks/guided-demos/notebook-ex-outputs/interactivetest.yaml
```diff
@@ -108,7 +108,6 @@ spec:
                       memory: 8G
                       nvidia.com/gpu: 0
                 imagePullSecrets: []
-          rayVersion: 2.1.0
```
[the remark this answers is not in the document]
Bobbins228: I am not entirely sure as of right now but I can add back the rayVersion in this PR and when I update the image again to Ray 2.23/CUDA 1.12 I can look into whether this is a breaking change for Job submission mode. WDYT?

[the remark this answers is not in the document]
astefanutti: Sounds good. I think it's safer that way as I think there is an explicit check in Kuberay when using the Job submission mode.

Bobbins228:
The e2e tests fail here due to the CertGenerator image on the CFO side when setting up the Kind Cluster.
Here is [proof](https://github.com/Bobbins228/codeflare-sdk/pull/5) of a passing e2e test with the new Ray image for the CertGenerator. 

I have made a [PR](https://github.com/project-codeflare/codeflare-operator/pull/567) to update the CFO with the new image

Srihari1192 (approved):

PR #531, opened 2024-05-09
simplify function calls and add option for custom resources

KPostOffice:
# Issue link

# What changes have been made
I changed some of the huge function calls that were getting out of hand to take a reference to the cluster object. It might be worthwhile to make these into class methods at a future date

I have also added the ability for the user to specify custom resource requirements for Ray
# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -29,6 +29,7 @@ from .auth import config_check, api_config_handler
 from ..utils import pretty_print
 from ..utils.generate_yaml import (
     generate_appwrapper,
+    head_worker_gpu_count_from_cluster,
 )
 from ..utils.kube_api_helpers import _kube_api_error_handling
 from ..utils.generate_yaml import is_openshift_cluster
@@ -118,48 +119,7 @@ class Cluster:
                     f"Namespace {self.config.namespace} is of type {type(self.config.namespace)}. Check your Kubernetes Authentication."
                 )
 
-        # Before attempting to create the cluster AW, let's evaluate the ClusterConfig
-
-        name = self.config.name
-        namespace = self.config.namespace
-        head_cpus = self.config.head_cpus
-        head_memory = self.config.head_memory
-        num_head_gpus = self.config.num_head_gpus
-        worker_cpu_requests = self.config.worker_cpu_requests
-        worker_cpu_limits = self.config.worker_cpu_limits
-        worker_memory_requests = self.config.worker_memory_requests
-        worker_memory_limits = self.config.worker_memory_limits
-        num_worker_gpus = self.config.num_worker_gpus
-        workers = self.config.num_workers
-        template = self.config.template
-        image = self.config.image
-        appwrapper = self.config.appwrapper
-        env = self.config.envs
-        image_pull_secrets = self.config.image_pull_secrets
-        write_to_file = self.config.write_to_file
-        local_queue = self.config.local_queue
-        labels = self.config.labels
-        return generate_appwrapper(
-            name=name,
-            namespace=namespace,
-            head_cpus=head_cpus,
-            head_memory=head_memory,
-            num_head_gpus=num_head_gpus,
-            worker_cpu_requests=worker_cpu_requests,
-            worker_cpu_limits=worker_cpu_limits,
-            worker_memory_requests=worker_memory_requests,
-            worker_memory_limits=worker_memory_limits,
-            num_worker_gpus=num_worker_gpus,
-            workers=workers,
-            template=template,
-            image=image,
-            appwrapper=appwrapper,
-            env=env,
-            image_pull_secrets=image_pull_secrets,
-            write_to_file=write_to_file,
-            local_queue=local_queue,
-            labels=labels,
-        )
+        return generate_appwrapper(self)
 
     # creates a new cluster with the provided or default spec
     def up(self):
@@ -305,7 +265,7 @@ class Cluster:
 
             if print_to_console:
                 # overriding the number of gpus with requested
-                cluster.worker_gpu = self.config.num_worker_gpus
+                _, cluster.worker_gpu = head_worker_gpu_count_from_cluster(self)
                 pretty_print.print_cluster_status(cluster)
         elif print_to_console:
             if status == CodeFlareClusterStatus.UNKNOWN:
@@ -443,6 +403,29 @@ class Cluster:
         """
         return self.job_client.get_job_logs(job_id)
 
+    @staticmethod
+    def _head_worker_extended_resources_from_rc_dict(rc: Dict) -> Tuple[dict, dict]:
+        head_extended_resources, worker_extended_resources = {}, {}
+        for resource in rc["spec"]["workerGroupSpecs"][0]["template"]["spec"][
+            "containers"
+        ][0]["resources"]["limits"].keys():
+            if resource in ["memory", "cpu"]:
+                continue
+            worker_extended_resources[resource] = rc["spec"]["workerGroupSpecs"][0][
+                "template"
+            ]["spec"]["containers"][0]["resources"]["limits"][resource]
+
+        for resource in rc["spec"]["headGroupSpec"]["template"]["spec"]["containers"][
+            0
+        ]["resources"]["limits"].keys():
+            if resource in ["memory", "cpu"]:
+                continue
+            head_extended_resources[resource] = rc["spec"]["headGroupSpec"]["template"][
+                "spec"
+            ]["containers"][0]["resources"]["limits"][resource]
+
+        return head_extended_resources, worker_extended_resources
+
     def from_k8_cluster_object(
         rc,
         appwrapper=True,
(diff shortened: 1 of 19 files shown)


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -875,6 +855,11 @@ def _map_to_ray_cluster(rc) -> Optional[RayCluster]:
                     protocol = "https"
             dashboard_url = f"{protocol}://{ingress.spec.rules[0].host}"
 
+    (
+        head_custom_resources,
```
Fiona-Waters, line 859: I am wondering about the use of `custom_resources` - I am not sure if it is clear here what will be expected from a user. To me custom resources are like RayClusters or LocalQueues. I understand that we want to avoid using just GPU as it is not broad enough. Could we use the term accelerators maybe?

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -460,6 +415,29 @@ def job_logs(self, job_id: str) -> str:
...
+            worker_custom_resources[resource] = rc["spec"]["workerGroupSpecs"][0][
+                "template"
+            ]["spec"]["containers"][0]["resources"]["limits"][resource]
+
+        for resource in rc["spec"]["headGroupSpec"]["template"]["spec"]["containers"][
+            0
+        ]["resources"]["limits"].keys():
+            if resource in ["memory", "cpu"]:
+                continue
+            head_custom_resources[resource] = rc["spec"]["headGroupSpec"][0][
```
Bobbins228, line 435:
```suggestion
            head_custom_resources[resource] = rc["spec"]["headGroupSpec"][
```

src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -120,64 +125,121 @@ def update_env(spec, env):
...
+
+def head_worker_gpu_count_from_cluster(
+    cluster: "codeflare_sdk.cluster.Cluster",
+) -> typing.Tuple[int, int]:
+    head_gpus = 0
+    worker_gpus = 0
+    for k in cluster.config.head_custom_resource_requests.keys():
+        resource_type = cluster.config.custom_resource_mapping[k]
+        if resource_type == "GPU":
+            head_gpus += cluster.config.head_custom_resource_requests[k]
```
Bobbins228, line 160: This function fails when running `get_cluster()` with accelerators specified with the below error.
```
 unsupported operand type(s) for +=: 'int' and 'str'
```
We could force the type conversion here

src/codeflare_sdk/cluster/config.py
```diff
@@ -63,6 +80,46 @@ def __post_init__(self):
...
+                )
+        self.custom_resource_mapping = {
+            **DEFAULT_RESOURCE_MAPPING,
+            **self.custom_resource_mapping,
+        }
+
+    def _gpu_to_resource(self):
+        if self.head_gpus:
+            warnings.warn(
+                "head_gpus is being deprecated, use head_custom_resource_requests"
```
Fiona-Waters, line 107: Maybe this warning message could also give an example of how to specify the resources since it isn't a simple changeover of the variable name but also a change in type.

src/codeflare_sdk/cluster/config.py
```diff
@@ -63,6 +80,46 @@ def __post_init__(self):
...
+            )
+            if "nvidia.com/gpu" in self.head_custom_resource_requests:
+                raise ValueError(
+                    "nvidia.com/gpu already exists in head_custom_resource_requests"
+                )
+            self.head_custom_resource_requests["nvidia.com/gpu"] = self.head_gpus
+        if self.num_gpus:
+            warnings.warn(
+                "num_gpus is being deprecated, use worker_custom_resource_requests"
```
Fiona-Waters, line 116: and the same here.

src/codeflare_sdk/cluster/cluster.py
KPostOffice: WDYT about `extended_resources` as defined here https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#consuming-extended-resources

Fiona-Waters: Yes I think that's better. We should also document this (when the notebooks are updated)  so that users know it is referring to accelerators. At a minimum with a comment beside the parameters in the notebook cluster config.

src/codeflare_sdk/cluster/cluster.py
```diff
@@ -456,6 +415,29 @@ def job_logs(self, job_id: str) -> str:
         """
         return self.job_client.get_job_logs(job_id)
 
+    @staticmethod
+    def _head_worker_extended_resources_from_rc_dict(rc: Dict) -> Tuple[dict, dict]:
+        head_extended_resources, worker_extended_resources = {}, {}
+        for resource in rc["spec"]["workerGroupSpecs"][0]["template"]["spec"][
+            "containers"
+        ][0]["resources"]["limits"].keys():
```
varshaprasad96, line 423: Just curious, is it intentional to only processes the first container and extract custom extended resources?

KPostOffice: Yeah, the extended resources will only apply to the container running the actual Ray runtime. The code naively assumes that this will always be the first listed container

src/codeflare_sdk/cluster/config.py
```diff
@@ -40,6 +79,7 @@ class ClusterConfiguration:
     head_memory: typing.Union[int, str] = 8
     head_gpus: int = None  # Deprecating
     num_head_gpus: int = 0
```
Bobbins228, line 81:
```suggestion
```
The `num_head_gpus` and `num_worker_gpus` should be removed.

Bobbins228 (changes requested):
The work surrounding `num_worker_gpus` and `num_head_gpus` should be removed in favour of `head_extended_resource_requests` and `worker_extended_resource_requests`

ChristianZaccaria (dismissed):
Tested and Job completed successfully with gpus.

Something to note, for some reason the ray dashboard believes we are using Tesla T4 GPUs. Does anyone know why?
image

I also noticed that while the pods are initialising, the status is SUSPENDED, eventhough pods are actually coming up. Moving from SUSPENDED->READY. Should I open an issue for both of these?

src/codeflare_sdk/cluster/config.py
```diff
@@ -69,9 +114,45 @@ def __post_init__(self):
...
+                    f"Resource mapping already exists for {overwritten}, set overwrite_default_resource_mapping to True to overwrite"
+                )
+        self.custom_resource_mapping = {
+            **DEFAULT_RESOURCE_MAPPING,
+            **self.custom_resource_mapping,
+        }
+
+    def _gpu_to_resource(self):
+        if self.num_head_gpus:
+            if "nvidia.com/gpu" in self.head_extended_resource_requests:
```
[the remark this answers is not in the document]
KPostOffice: This would happen if a user specified:
ClusterConfig(num_gpus=1, worker_extended_resources={"nvidia.com/gpu": 1}).

Just preventing the user from trying to specify the num gpus twice until the num_gpus arg is fully deprecated

src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -113,67 +118,111 @@ def update_resources(
...
+            requests[k] = custom_resources[k]
+
+
+def head_worker_gpu_count_from_cluster(
+    cluster: "codeflare_sdk.cluster.Cluster",
+) -> typing.Tuple[int, int]:
+    head_gpus = 0
+    worker_gpus = 0
+    for k in cluster.config.head_extended_resource_requests.keys():
+        resource_type = cluster.config.custom_resource_mapping[k]
```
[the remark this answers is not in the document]
KPostOffice: I can add a warning to check and make sure that all extended resources exist in the resource mapping and error out in config

src/codeflare_sdk/cluster/config.py
```diff
@@ -70,8 +113,60 @@ def __post_init__(self):
...
+        self, extended_resources: typing.Dict[str, int]
+    ):
+        for k in extended_resources.keys():
+            if k not in self.extended_resource_mapping.keys():
+                raise ValueError(
+                    f"extended resource '{k}' not found in extended_resource_mapping, available resources are {list(self.extended_resource_mapping.keys())}, to add more supported resources use extended_resource_mapping. i.e. extended_resource_mapping = {{'{k}': 'FOO_BAR'}}"
+                )
+
+    def _gpu_to_resource(self):
+        if self.head_gpus:
```
Bobbins228, line 152: There is a case here where I specified `num_gpus`, `head_gpus` and `head_extended_resource_requests={"amd.com/gpu": 1}`, `worker_extended_resource_requests={"amd.com/gpu": 2}`

The AMD GPU was overwritten by the `nvidia.com` GPU. Should we concern ourselves with this case or trust that the user will read the warning and remove head/num_gpus?

KPostOffice: It shouldn't overwrite, it would end up specifying both `nvidia.com/gpu=num_gpus` and `amd.com/gpu=2` in the extended resource dict. I was having head_gpus just add the nvidia_gpus in order to match previous functionality since we assumed it was nvidia by default before.

Bobbins228: I'm for appending the the accelerators with the default nvidia gpus for now 👍

Bobbins228 (approved):
/approve

PR #532, opened 2024-05-09
RHOAIENG-5344 - E2E test for Ray local interactive

jiripetrlik:
# Issue link
https://issues.redhat.com/browse/RHOAIENG-5344

### DIFF
--- a/tests/e2e/local_interactive_sdk_oauth_test.py
+++ b/tests/e2e/local_interactive_sdk_oauth_test.py
@@ -1,0 +1,87 @@
+from codeflare_sdk import (
+    Cluster,
+    ClusterConfiguration,
+    TokenAuthentication,
+    generate_cert,
+)
+
+import math
+import pytest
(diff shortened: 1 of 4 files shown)


tests/e2e/local_interactive_sdk_oauth_test.py
```diff
@@ -0,0 +1,73 @@
+from codeflare_sdk import Cluster, ClusterConfiguration, TokenAuthentication, generate_cert
+
+import math
+import pytest
+import ray
+
+from support import *
+
+@pytest.mark.openshift
+class TestRayLocalInteractiveOauth:
```
sutaakar, line 10: Would it be possible to create another test without oauth to be executed against KinD?

Srihari1192 (approved):
LGTM

one nit pick , its good to remove created kueue resources aswell, i see its not handled in other tests too

PR #533, opened 2024-05-10
Adding validation for local_queue provided in cluster config

Fiona-Waters:
# Issue link
[RHOAIENG-7064](https://issues.redhat.com/browse/RHOAIENG-7064)

# What changes have been made
Currently when specifying local_queue, if any value is provided the ray cluster can be created. Added a check to validate local_queue name provided by user in cluster configuration exists.

# Verification steps
 - Install RHOAI with codeflare components set to Managed
 - In a jupyter notebook install codeflare sdk from this branch
 ```
git clone https://github.com/Fiona-Waters/codeflare-sdk.git -b local-queue
cd codeflare-sdk
pip install poetry
poetry install
```
 - Restart the kernal and run through a demo notebook
   - Ensure that you do not have the required annotation for the default local queue set.
- scenario 1
   - Set the local_queue parameter to `foobar` and check that you cannot continue and you see the correct error message.
- scenario 2
   - Provide an existing local_queue name and see that you can continue and create the ray cluster
- scenario 3
   - Comment out the local_queue parameter and see that you get the correct error message.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests

### DIFF
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -308,6 +308,26 @@ def get_default_kueue_name(namespace: str):
     )
 
 
+def local_queue_exists(namespace: str, local_queue_name: str):
+    # get all local queues in the namespace
+    try:
+        config_check()
+        api_instance = client.CustomObjectsApi(api_config_handler())
+        local_queues = api_instance.list_namespaced_custom_object(
+            group="kueue.x-k8s.io",
+            version="v1beta1",
+            namespace=namespace,
+            plural="localqueues",
+        )
+    except Exception as e:  # pragma: no cover
+        return _kube_api_error_handling(e)
+    # check if local queue with the name provided in cluster config exists
+    for lq in local_queues["items"]:
+        if lq["metadata"]["name"] == local_queue_name:
+            return True
+    return False
+
+
 def write_components(
     user_yaml: dict,
     output_file_name: str,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/codeflare_sdk/utils/generate_yaml.py [hunk 1]
```diff
@@ -308,6 +308,26 @@ def get_default_kueue_name(namespace: str):
     )
 
 
+def local_queue_exists(namespace, local_queue_name: str):
```
franciscojavierarceo, line 311: is namespace also a str? probably good to add that type hint explicitly.

franciscojavierarceo, line 311:
```suggestion
def local_queue_exists(namespace: str, local_queue_name: str) -> bool:
```

franciscojavierarceo (approved):

src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -355,26 +380,31 @@ def load_components(
...
     lq_name = local_queue or get_default_kueue_name(namespace)
     cluster_labels = labels
-    for component in components:
-        if "generictemplate" in component:
-            if (
-                "workload.codeflare.dev/appwrapper"
-                in component["generictemplate"]["metadata"]["labels"]
-            ):
-                del component["generictemplate"]["metadata"]["labels"][
+    if local_queue_exists(namespace, lq_name):
```
KPostOffice, line 383: small nit, can you change this to
```suggestion
    if not local_queue_exists(namespace, lq_name):
        raise ...
   // unindented default block  
   ...
```

Fiona-Waters: Updated.

KPostOffice: It looks the same to me.

Fiona-Waters: Apologies, I don't know what happened there. It is now updated.

PR #534, opened 2024-05-10
Add warning if Kueue CRDs not installed on cluster

oksanabaza:
# Issue link

[RHOAIENG-6294](https://issues.redhat.com/browse/RHOAIENG-6294)
# What changes have been made

A warning has been added to the ClusterConfiguration cell to inform users if Kueue CRDs are not installed on the cluster and local_queue is not provided. In this scenario, users will receive informative logs indicating that the absence of CRDs may result in missing functionality.
# Verification steps

Uninstall Kueue and delete any CustomResourceDefinitions (CRDs) that exist (filter for `CustomResourceDefinition` in resources and select `LocalQueue` in the name). Clone this branch into Jupyter Notebook, attempt to create a Ray cluster. In the ClusterConfiguration cell, comment out the `local_queue="local-queue-name"` line. When you run it, you should receive a message indicating that `Kueue is not installed or won't be used. The absence of CRDs may result in missing functionality.`
## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -321,6 +321,10 @@ def write_components(
         os.makedirs(directory_path)
 
     components = user_yaml.get("spec", "resources")["resources"].get("GenericItems")
+    if local_queue is None:
+        print(
+            "Kueue is not installed or won't be used. The absence of CRDs may lack the necessary functionality."
+        )
     open(output_file_name, "w").close()
     lq_name = local_queue or get_default_kueue_name(namespace)
     cluster_labels = labels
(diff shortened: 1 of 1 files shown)


src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -321,6 +321,10 @@ def write_components(
         os.makedirs(directory_path)
 
     components = user_yaml.get("spec", "resources")["resources"].get("GenericItems")
+    if local_queue is None:
```
Fiona-Waters, line 324: This will only check if the local_queue variable has been set by the user in the cluster config. If we want to check for the existence of the queue CRDs we could do it like it is done [here](https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/cluster/cluster.py#L258) with the RayCluster CRD.

oksanabaza:
Closing this pull request as the previous changes cover this issue

PR #538, opened 2024-05-16
Rhoaieng 7141

Ygnas:
# Issue link

[RHOAIENG-7141](https://issues.redhat.com/browse/RHOAIENG-7141)
# What changes have been made

`get_cluster` should not fail if there is no default local_queue in `RayCluster`
# Verification steps

Start a new `RayCluster` with a LocalQueue specified and no default one, and run:

`
get_cluster("raytest")
`

Should not got error:
`
Default Local Queue with kueue.x-k8s.io/default-queue: true annotation not found please create a default Local Queue or provide the local_queue name in Cluster Configuration`
## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -541,6 +541,7 @@ def from_k8_cluster_object(
             mcad=mcad,
             write_to_file=write_to_file,
             verify_tls=verify_tls,
+            local_queue=rc["metadata"]["labels"]["kueue.x-k8s.io/queue-name"],
```
KPostOffice, line 544:
```suggestion
            local_queue=rc["metadata"].get("labels", dict()).get("kueue.x-k8s.io/queue-name", None),
```

Ygnas: Thanks will update it tomorrow, but can a `RayCluster` even exists without `local_queue`? I thought you can't create one without it so label should always exists?

KPostOffice: Not via the SDK for sure, but I'm not positive we can guarantee that every RayCluster has the label. I would need to think on it. I think it's better to be safe than sorry here mainly.

KPostOffice (approved):

PR #542, opened 2024-05-21
Use native gRPC resolver in PR check for local interactive

sutaakar:
This resolver is compatible with dnsmasq used for dynamic hostnames in KinD.

# Issue link

https://issues.redhat.com/browse/RHOAIENG-7055

# What changes have been made

Use native gRPC resolver which is compatible with dnsmasq.
Use Python 3.9 which is compatible with latest Ray image in CodeFlare repo.

# Verification steps

Should be verified together with changes in https://github.com/project-codeflare/codeflare-sdk/pull/532

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -76,6 +76,12 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Set up specific Python version
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+          cache: 'pip' # caching pip dependencies
+
       - name: Setup and start KinD cluster
         uses: ./common/github-actions/kind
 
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_tests.yaml
```diff
@@ -114,6 +120,8 @@ jobs:
           kubectl create clusterrolebinding sdk-user-clusterqueue-creator --clusterrole=clusterqueue-creator --user=sdk-user
           kubectl create clusterrole localqueue-creator --verb=get,list,create,delete,patch --resource=localqueues
           kubectl create clusterrolebinding sdk-user-localqueue-creator --clusterrole=localqueue-creator --user=sdk-user
+          kubectl create clusterrole list-secrets --verb=get,list --resource=secrets
+          kubectl create clusterrolebinding sdk-user-list-secrets --clusterrole=list-secrets --user=sdk-user
```
astefanutti, line 124: We would need to review the entire set of RBAC that are created there, but right now that's OK.

jiripetrlik (approved):

PR #539, opened 2024-05-16
[WIP] Fixed get_current_namespace for in-cluster configuration

Bobbins228:
# Issue link

# What changes have been made

# Verification steps

## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #543, opened 2024-05-21
SA get current namespace

KPostOffice:
# Issue link
supersedes #539 

# What changes have been made
add check for SA namespace as default

# Verification steps
use RHOAI workbook, clone this branch and `run pip install .` see that the namespace retrieved is the same as the project that you are in

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/cluster.py
+++ b/src/codeflare_sdk/cluster/cluster.py
@@ -630,14 +630,29 @@ def get_current_namespace():  # pragma: no cover
             print("Unable to find current namespace")
             return None
     else:
-        try:
-            _, active_context = config.list_kube_config_contexts(config_check())
-        except Exception as e:
-            return _kube_api_error_handling(e)
-        try:
-            return active_context["context"]["namespace"]
-        except KeyError:
-            return None
+        if "KUBERNETES_PORT" in os.environ:
+            if os.path.isfile(
+                "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+            ):
+                try:
+                    file = open(
+                        "/var/run/secrets/kubernetes.io/serviceaccount/namespace", "r"
+                    )
+                    active_context = file.readline().strip("\n")
+                    return active_context
+                except Exception as e:
+                    print(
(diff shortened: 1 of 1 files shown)


ChristianZaccaria (commented):
Will we want to enable the following test as part of this fix, or address in a separate PR?
https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/cluster/cluster.py#L724 

Apart from the above question, all works as expected!
https://github.com/project-codeflare/codeflare-sdk/blob/main/tests/unit_test.py#L579

ChristianZaccaria:
Wondering if it's a good idea to set something like this to all notebooks, further removing the need for user input:
```
    name='raytest',
    namespace= cluster.get_current_namespace(),
```
I think we would need to add the import too for usage:
`from codeflare_sdk.cluster import cluster`

KPostOffice:
> Wondering if it's a good idea to set something like this to all notebooks, further removing the need for user input:
> 
> ```
> cluster = Cluster(ClusterConfiguration(
>     name='raytest',
>     namespace= cluster.get_current_namespace(),
> ```
> 
> I think we would need to add the import too for usage: `from codeflare_sdk.cluster import cluster`

If config doesn't have a namespace set [it should use the current working namespace](https://github.com/project-codeflare/codeflare-sdk/blob/b4c55e4910222e2be8a7f8c7026b98287c9a9cbe/src/codeflare_sdk/cluster/cluster.py#L143)

KPostOffice:
> Will we want to enable the following test as part of this fix, or address in a separate PR? https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/cluster/cluster.py#L724
> 
> https://github.com/project-codeflare/codeflare-sdk/blob/main/tests/unit_test.py#L579
> 
> Apart from the above question, all works as expected!

I'm not sure what the test is testing exactly. Can we leave it out for now?

Fiona-Waters:
I ran a demo notebook without providing the namespace and the `rhods-notebooks` namespace was retrieved from `/var/run/secrets/kubernetes.io/serviceaccount/namespace`. Works as expected.
/approve

PR #489, opened 2024-03-22
Made ODH cert default cert for Token Auth

Bobbins228:
# Issue link

[RHOAIENG-52](https://issues.redhat.com/browse/RHOAIENG-52), [RHOAIENG-4375](https://issues.redhat.com/browse/RHOAIENG-4375)
# What changes have been made

For `TokenAuthentication` the SDK will use the cert injected into a ODH/RHOAI Notebook by default in the `/etc/pki/tls/custom-certs/ca-bundle.crt` location
# Verification steps

### Required steps for ODH/RHOAI
* ODH > 2.9
* Create a Data Science Workbench
* Ensure that the path `/etc/pki/tls/custom-certs/ca-bundle.crt` exists in the notebook.
### Notebook server ODH
* Clone this repository with `git clone https://github.com/project-codeflare/codeflare-sdk.git`
* Checkout this PR's branch
* Run `poetry build` - install if needed (`pip install poetry`)
* Run `pip install --force-reinstall dist/codeflare_sdk-0.0.0.dev0-py3-none-any.whl` 
* Restart your notebook kernel
## Testing
Authenticate with `TokenAuthentication`.
Set the environment variable `CA_CERT_PATH` to the path to your cert before authenticating 
```
auth = TokenAuthentication(
    token = "XXXXX",
    server = "XXXXX",
    skip_tls=False
)
auth.login()
```
You should receive the message `Authenticated with certificate located at /etc/pki/tls/custom-certs/ca-bundle.crt`
You should be able to run through any demo notebook.
## Checks
- [x] I've made sure the tests are passing. 
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] Testing is not required for this change

PR #544, opened 2024-05-21
Workbench certs

KPostOffice:
# Issue link
supersedes #489 

# What changes have been made
I made a minor change to #489 since @Bobbins228 is out. This change makes sure the explicitly passed arg takes priority over the default

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


src/codeflare_sdk/cluster/auth.py
```diff
@@ -106,10 +118,20 @@ def login(self) -> str:
             configuration.api_key_prefix["authorization"] = "Bearer"
             configuration.host = self.server
             configuration.api_key["authorization"] = self.token
-            if self.skip_tls == False and self.ca_cert_path == None:
+
+            if self.skip_tls == False:
```
ChristianZaccaria, line 122:
```suggestion
            if not self.skip_tls:
```
Nitpick for a more Pythonic style.

src/codeflare_sdk/cluster/auth.py
```diff
@@ -106,10 +118,20 @@ def login(self) -> str:
...
+                if self.ca_cert_path == None:
```
ChristianZaccaria, line 123:
```suggestion
                if self.ca_cert_path is None:
```

tests/unit_test.py
```diff
@@ -131,19 +131,24 @@ def test_token_auth_creation():
...
         assert token_auth.ca_cert_path == None
 
+        os.environ["CF_SDK_CA_CERT_PATH"] = f"/etc/pki/tls/custom-certs/ca-bundle.crt"
         token_auth = TokenAuthentication(token="token", server="server", skip_tls=False)
         assert token_auth.token == "token"
         assert token_auth.server == "server"
         assert token_auth.skip_tls == False
-        assert token_auth.ca_cert_path == None
+        assert token_auth.ca_cert_path == "/etc/pki/tls/custom-certs/ca-bundle.crt"
+        os.environ.pop("CF_SDK_CA_CERT_PATH")
```
ChristianZaccaria, line 140: I think this line is not necessary, further proving that priority is given to the `ca_cert_path` parameter even if the environment variable is set.

ChristianZaccaria (changes requested):
Tested:
- Priority is given to the `ca_cert_path` argument over the environment variable set.
- Can authenticate with certificate located through the path specified by the environment variable.
- Ran through a notebook and submitted job successfully.

Thank you, just left a few nitpicks.

tests/unit_test.py
KPostOffice: It was failing some other unit tests since they don't expect the env var to be set

ChristianZaccaria (approved):

PR #550, opened 2024-05-31
Skip namespace deletion for upgrade testing

Srihari1192:
# Issue link

https://issues.redhat.com/browse/RHOAIENG-7032

# What changes have been made

Removed namespace deletion in **TestMnistJobSubmit** as after post upgrade job submission needs to verify in Distributed workload  resource usages in UI

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/tests/upgrade/raycluster_sdk_upgrade_test.py
+++ b/tests/upgrade/raycluster_sdk_upgrade_test.py
@@ -80,9 +80,6 @@ class TestMnistJobSubmit:
         if not self.cluster:
             raise RuntimeError("TestRayClusterUp needs to be run before this test")
 
-    def teardown_method(self):
-        delete_namespace(self)
-
     def test_mnist_job_submission(self):
         auth = TokenAuthentication(
             token=run_oc_command(["whoami", "--show-token=true"]),


sutaakar:
Who/what code will be responsible for deleting the namespace?

sutaakar:
What is the use case?
To start RayCluster, do the upgrade and after upgrade to verify that UI shows correct values?

If it is so, can you use same approach as in https://github.com/red-hat-data-services/ods-ci/pull/1467 ?

Srihari1192:
> What is the use case? To start RayCluster, do the upgrade and after upgrade to verify that UI shows correct values?
> 
> If it is so, can you use same approach as in [red-hat-data-services/ods-ci#1467](https://github.com/red-hat-data-services/ods-ci/pull/1467) ?

The Use case which i am implementing is 

- In pre-upgrade scenario Run [TestMNISTRayClusterUp]( https://github.com/project-codeflare/codeflare-sdk/blob/2a43d9e05f750fd7f5ef14ef2e032c75b51c569b/tests/upgrade/raycluster_sdk_upgrade_test.py#L16)  verify Ray cluster up scenario and check resource usage in UI that showing correct values
- In post upgrade scenario Run [TestMnistJobSubmit](https://github.com/project-codeflare/codeflare-sdk/blob/main/tests/upgrade/raycluster_sdk_upgrade_test.py#L75) and verify Ray cluster exists, submit job, monitor  job submission status and check resource usage in UI  that showing correct values

Yes we can use this approach [red-hat-data-services/ods-ci#1467](https://github.com/red-hat-data-services/ods-ci/pull/1467) to create ray cluster but we may miss coverage of actual user usage perspective of ray cluster creation and job submission wdyt?

sutaakar:
hmm, I guess there is no easier way to achieve it
fine for me

sutaakar (approved):

abhijeet-dhumal (approved):

PR #554, opened 2024-06-06
Added Custom Volume/Volume Mount Support

Bobbins228:
# Issue link

[RHOAIENG-8262](https://issues.redhat.com/browse/RHOAIENG-8262)
# What changes have been made

Added a two new config variables `volumes` which is used for adding custom volumes on a Ray Cluster and `volume_mounts` which does the same but for VolumeMounts.
# Verification steps

* Checkout this PR's branch
## Testing
The `volumes` and `volume_mounts` config parameters are both lists that take the `V1Volume` and `V1VolumeMount` objects respectively.
Below I have given some examples for testing purposes.
### Volumes/Volume Mounts configuration
```
from kubernetes.client import V1Volume, V1VolumeMount, V1EmptyDirVolumeSource, V1ConfigMapVolumeSource, V1KeyToPath, V1SecretVolumeSource
volume_mounts = [
    V1VolumeMount(
        mount_path="/home/ray/test1",
        name = "test"
    ),
    V1VolumeMount(
        mount_path = "/home/ray/test2",
        name = "test2",
    ),
    V1VolumeMount(
        mount_path = "/home/ray/test3",
        name = "test3",
    )
]

volumes = [
    V1Volume(
        name="test",
        empty_dir=V1EmptyDirVolumeSource(size_limit="2Gi"),
    ),
    V1Volume(
        name="test2",
        config_map=V1ConfigMapVolumeSource(
            name="test-config-map",
            items=[V1KeyToPath(key="test", path="data.txt")]
        )
    ),
    V1Volume(
        name="test3",
        secret=V1SecretVolumeSource(
            secret_name="test-secret"
    )
)
]
```
### ClusterConfiguration
```
    name='raytest',
    namespace='default', # Update to your namespace
    num_workers=1,
    min_cpus=2,
    max_cpus=2,
    head_memory=3,
    min_memory=3,
    max_memory=3,
    num_gpus=0,
    head_gpus=0,
    image="quay.io/project-codeflare/ray:latest-py39-cu118",
    write_to_file=True, # When enabled Ray Cluster yaml files are written to /HOME/.codeflare/resources 
    volume_mounts = volume_mounts,
    volumes = volumes
))
```
### Test Config Map
```
kind: ConfigMap
apiVersion: v1
metadata:
  name: test-config-map
  namespace: YOUR_NAMESPACE
immutable: false
data:
  test: pass
```
### Test Secret
* Create a secret on OpenShift by going to Workloads -> Secrets -> Create -> key/vaule secret
* Set the name to ` test-secret` 
* The key/value can be whatever you like

### Verify
The volumes & volume mounts should be added to your worker and head pods.
You can exec into your pod and check that the emptyDir, Config Map volume and Secret Volume were mounted correctly
## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/config.py
+++ b/src/codeflare_sdk/cluster/config.py
@@ -53,6 +53,8 @@ class ClusterConfiguration:
     write_to_file: bool = False
     verify_tls: bool = True
     labels: dict = field(default_factory=dict)
+    volumes: list = field(default_factory=list)
+    volume_mounts: list = field(default_factory=list)
 
     def __post_init__(self):
         if not self.verify_tls:
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -96,6 +96,20 @@ def update_image_pull_secrets(spec, image_pull_secrets):
     ]
 
 
+def update_volume_mounts(spec, volume_mounts: list):
+    containers = spec.get("containers")
+    for volume_mount in volume_mounts:
+        for container in containers:
+            volumeMount = client.ApiClient().sanitize_for_serialization(volume_mount)
+            container["volumeMounts"].append(volumeMount)
+
+
+def update_volumes(spec, volumes: list):
+    for volume in volumes:
+        new_volume = client.ApiClient().sanitize_for_serialization(volume)
+        spec["volumes"].append(new_volume)
+
+
 def update_env(spec, env):
     containers = spec.get("containers")
     for container in containers:
@@ -136,6 +150,8 @@ def update_nodes(
     head_cpus,
     head_memory,
     head_gpus,
+    volumes,
+    volume_mounts,
 ):
     head = cluster_yaml.get("spec").get("headGroupSpec")
     head["rayStartParams"]["num-gpus"] = str(int(head_gpus))
@@ -150,6 +166,8 @@ def update_nodes(
 
     for comp in [head, worker]:
         spec = comp.get("template").get("spec")
+        update_volume_mounts(spec, volume_mounts)
+        update_volumes(spec, volumes)
         update_image_pull_secrets(spec, image_pull_secrets)
         update_image(spec, image)
         update_env(spec, env)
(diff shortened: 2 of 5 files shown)


src/codeflare_sdk/cluster/config.py
```diff
@@ -53,6 +53,7 @@ class ClusterConfiguration:
     write_to_file: bool = False
     verify_tls: bool = True
     labels: dict = field(default_factory=dict)
+    persistent_volumes: dict = field(default_factory=list)
```
KPostOffice, line 56: Could this directly be passed as a two args, `persistent_volume_claims` and `volume_mounts`, and use k8 python objects directly? (`V1Volume` and `V1VolumeMount`) My gut feeling is that this would save some headache in the future when more features are asked for around volumes (i.e. mounting secrets). If we do this it puts a bit more onus on the user, but most users should be fairly familiar with the python k8 api. WDYT?

KPostOffice: Yeah, I think that would be good. I would basically just go based off of this documentation. https://kubernetes.io/docs/concepts/storage/volumes. I guess it makes more sense for it to be `volumes` and `volume_mounts` now that I think about it a bit more since pvc is just a particular type of volume that you can specify

src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -99,6 +99,26 @@ def update_image_pull_secrets(spec, image_pull_secrets):
     ]
 
 
+def update_volume_mounts(spec, volume_mounts: list):
+    if volume_mounts is []:
+        return None
```
KPostOffice, line 104: This can be removed, the `for` loop won't execute anything if the list is empty. Here an below.

KPostOffice (commented):
Overall LGTM. One tiny comment

KPostOffice:
Conflict but then ready to come in I think

KPostOffice (approved):
/lgtm

PR #570, opened 2024-06-21
Add arguments to pass Ray cluster head and worker templates

### LINKED ISSUE #413: ClusterConfiguration should support tolerations
### ClusterConfiguration should support tolerations

Run Ray clusters (especially the worker pods) on tainted nodes. 

### Description of Problem the Feature Should Solve

You cannot create a Ray cluster with tolerations using the CodeFlare SDK
cluster = Cluster(ClusterConfiguration(...))

Often, machine nodes are tainted to prevent unwanted workloads.  This is especially the case in GPU nodes which are often tainted.  In addition different nodes will have different sized gpus, which would also use taints to make sure the correct workers land on the correct nodes.

You might also want to add a toleration to the `headGroupSpec`

### Describe the Solution You Would Like to See

Add `worker_tolerations` and `head_tolderations` as optional parameters for `ClusterConfiguration`
```
    head_tolerations=[key, operator, effect],
    worker_tolerations=[key, operator, effect]
))
```

### Describe Alternatives You Have Considered

Editing the yaml file and just using kuberay directly

astefanutti:
# Issue link

Fixes #413.

# What changes have been made

This PR adds parameters to the Ray cluster configuration API, so users can provide Pod template for the head and worker nodes, e.g.:

```python
from kubernetes import V1PodTemplateSpec, V1PodSpec, V1Toleration

    worker_template=V1PodTemplateSpec(
        spec=V1PodSpec(
            tolerations=[V1Toleration(
                key="nvidia.com/gpu",
                operator="Exists",
                effect="NoSchedule",
            )],
            node_selector={
                "nvidia.com/gpu.present": "true",
            },
        )
    ),
    head_template=V1PodTemplateSpec(...),
))
```

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -20,16 +20,12 @@ This sub-module exists primarily to be used internally by the Cluster object
 from typing import Optional
 import typing
 import yaml
-import sys
 import os
-import argparse
 import uuid
 from kubernetes import client, config
 from .kube_api_helpers import _kube_api_error_handling
 from ..cluster.auth import api_config_handler, config_check
-from os import urandom
-from base64 import b64encode
-from urllib3.util import parse_url
+from mergedeep import merge, Strategy
 
 
 def read_template(template):
@@ -278,6 +274,16 @@ def write_user_yaml(user_yaml, output_file_name):
     print(f"Written to: {output_file_name}")
 
 
+def apply_head_template(cluster_yaml: dict, head_template: client.V1PodTemplateSpec):
+    head = cluster_yaml.get("spec").get("headGroupSpec")
+    merge(head["template"], head_template.to_dict(), strategy=Strategy.ADDITIVE)
+
+
+def apply_worker_template(cluster_yaml: dict, worker_template: client.V1PodTemplateSpec):
+    worker = cluster_yaml.get("spec").get("workerGroupSpecs")[0]
+    merge(worker["template"], worker_template.to_dict(), strategy=Strategy.ADDITIVE)
+
+
 def generate_appwrapper(
     name: str,
     namespace: str,
@@ -291,6 +297,8 @@ def generate_appwrapper(
     gpu: int,
     workers: int,
     template: str,
+    head_template: client.V1PodTemplateSpec,
+    worker_template: client.V1PodTemplateSpec,
     image: str,
     appwrapper: bool,
     env,
@@ -302,6 +310,12 @@ def generate_appwrapper(
     volume_mounts: list[client.V1VolumeMount],
 ):
     cluster_yaml = read_template(template)
+
+    if head_template:
+        apply_head_template(cluster_yaml, head_template)
+    if worker_template:
+        apply_worker_template(cluster_yaml, worker_template)
+
     appwrapper_name, cluster_name = gen_names(name)
     update_names(cluster_yaml, cluster_name, namespace)
     update_nodes(
--- a/poetry.lock
+++ b/poetry.lock
@@ -1274,6 +1274,17 @@ lint = ["black", "check-manifest", "flake8", "isort", "mypy"]
 test = ["Cython", "greenlet", "ipython", "pytest", "pytest-cov", "setuptools"]
 
 [[package]]
+name = "mergedeep"
+version = "1.3.4"
+description = "A deep merge function for 🐍."
+optional = false
+python-versions = ">=3.6"
+files = [
+    {file = "mergedeep-1.3.4-py3-none-any.whl", hash = "sha256:70775750742b25c0d8f36c55aed03d24c3384d17c951b3175d898bd778ef0307"},
+    {file = "mergedeep-1.3.4.tar.gz", hash = "sha256:0096d52e9dad9939c3d975a774666af186eda617e6ca84df4c94dec30004f2a8"},
+]
+
+[[package]]
 name = "msgpack"
 version = "1.0.8"
 description = "MessagePack serializer"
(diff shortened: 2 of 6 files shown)


src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -278,6 +274,16 @@ def write_user_yaml(user_yaml, output_file_name):
...
 
 
+def apply_head_template(cluster_yaml: dict, head_template: client.V1PodTemplateSpec):
+    head = cluster_yaml.get("spec").get("headGroupSpec")
+    merge(head["template"], head_template.to_dict(), strategy=Strategy.ADDITIVE)
+
+
+def apply_worker_template(cluster_yaml: dict, worker_template: client.V1PodTemplateSpec):
+    worker = cluster_yaml.get("spec").get("workerGroupSpecs")[0]
+    merge(worker["template"], worker_template.to_dict(), strategy=Strategy.ADDITIVE)
```
KPostOffice, line 284: How should a user edit the container spec? Specifying partial container spec will only add a new partial entry to the `containers` list. This becomes an issue when specifying something like extra volume mounts.

KPostOffice (commented):
I have one comment, which pertains to the scope of what the user can provide. I think it might require more in depth discussion.

src/codeflare_sdk/utils/generate_yaml.py
astefanutti: Right, ideally we'd want to leverage strategic merge patch for that: https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch. It's easy to do in Go by using https://github.com/kubernetes/apimachinery/blob/master/pkg/util/strategicpatch/patch.go, but it don't know if that's possible in Python.

With `mergedeep` and the `ADDITIVE` strategy, lists are not replaced, but elements appended. The problem being that something like:
```python
        head_template=V1PodTemplateSpec(
          spec=V1PodSpec(
              containers=[
                  V1Container(name="ray-head", volume_mounts=[
                      V1VolumeMount(name="config", mount_path="path"),
                  ])
              ],
              volumes=[
                  V1Volume(name="config", config_map=V1ConfigMapVolumeSource(name="config")),
              ],
          )
        ),
```
It appends an entire new container, while it should only append the extra volume mount to the existing container.

Strategic merge patch solves this, by relying on information like `x-kubernetes-patch-merge-key` so it knows how to match items by key.

KPostOffice: I couldn't seem to find anything for merging V1PodTemplate specs, only namespaced pods by sending API requests. There's this package https://pypi.org/project/jsonmerge/ which we could use, it would require maintaining a bit of redundant config however 😒.

astefanutti: Yes, I've actually stumbled upon `jsonmerge` this morning as well. it seems it'd be possible to provide a merge strategy by key for lists. Let's give it a try? I agree we'd have to maintain some config, unless we figure a way to leverage Kubernetes JSON schema https://github.com/yannh/kubernetes-json-schema?tab=readme-ov-file, that do contain `x-kubernetes-patch-merge-key` information.

astefanutti: @varshaprasad96 maybe you would have some ideas?

varshaprasad96: One step better than just additive merging could be calculating the diff and then merging them ([deepdiff](https://github.com/seperman/deepdiff) and then [deepmerge](https://pypi.org/project/deepmerge/) - with probably custom merging strategies if needed). But this would still not solve the problem with conflicts at the very least. Looks like if we want to leverage JSON schema we either need to use a live client or load the config to guide merging process.

KPostOffice: Here's a script that we could use as a starting point for generate a jsonmerge schema from a k8 schema. I haven't gotten around to testing it yet. The output looks right based on my understanding of the `jsonmerge` package documentation, not based on testing done with actual objects

```
def merge_schema_from_k8_schema(k8_schema):
    to_return = {}
    k8_properties = {}
    if "array" in k8_schema.get("type", []) and k8_schema.get("x-kubernetes-list-type") != "atomic":
        to_return["items"] = {}
        to_return["items"]["properties"] = {}
        toret_properties = to_return["items"]["properties"]
        k8_properties = k8_schema.get("items", {}).get("properties", {})
    elif "object" in k8_schema.get("type", []):
        to_return["properties"] = {}
        toret_properties = to_return["properties"]
        k8_properties = k8_schema.get("properties", {})
    for key, value in k8_properties.items():
        if "object" in value.get("type", []) and "properties" in value:
            toret_properties[key] = merge_schema_from_k8_schema(value)
        elif "array" in value.get("type", []) and "items" in value:
            toret_properties[key] = merge_schema_from_k8_schema(value)
        else:
            toret_properties[key] = {"mergeStrategy": "overwrite"}

    if "array" in k8_schema.get("type", []):
        if k8_schema.get("x-kubernetes-list-type") == "set":
            to_return["mergeStrategy"] = "arrayMergeById"
            to_return["idRef"] = "/"
        elif k8_schema.get("x-kubernetes-list-type") == "atomic":
            to_return["mergeStrategy"] = "overwrite"
        elif k8_schema.get("x-kubernetes-list-type") == "map":
            if k8_schema.get("x-kubernetes-patch-merge-key"):
                to_return["mergeStrategy"] = "arrayMergeById"
                to_return["id"] = k8_schema["x-kubernetes-patch-merge-key"]
            else:
                to_return["mergeStrategy"] = "overwrite"
    elif "object" in k8_schema.get("type", []):
        to_return["mergeStrategy"] = "objectMerge"
    else:
        to_return["mergeStrategy"] = "overwrite"
    return to_return
```

Issue #504, opened 2024-04-09
Add support for __version__


diegolovison:
### Name of Feature or Improvement
As a developer I would like to know in runtime what is the version of `codeflare_sdk`
Having `__version__` is common for most of Python libraries

PR #575, opened 2024-06-25
Added support for checking version of the codeflare-sdk package

snalawad:
- Resolves: Issue #504 

# Issue link
Add support for __version__ : https://github.com/project-codeflare/codeflare-sdk/issues/504

# What changes have been made
Added support to check the version of the codeflare-sdk package.
Using `print(codeflare_sdk.__version__)` gives the version number of the version that is installed.

# Verification steps
Verified the usage as follows:

Screenshot from 2024-06-25 21-53-37
Screenshot from 2024-06-25 21-54-09

## Checks

- Testing Strategy
   - [ ] Tested the usage


src/codeflare_sdk/__init__.py
```diff
@@ -19,3 +19,13 @@
 from .job import RayJobClient
 
 from .utils import generate_cert
+
+from importlib.metadata import version, PackageNotFoundError
+
+try:
+    __version__ = version(
+        "codeflare-sdk"
+    )  # Update with latest version with each release
```
ChristianZaccaria, line 28: Just one question, wondering what do you mean by this comment?

ChristianZaccaria (commented):
Thanks for the PR! - Tested and works as described.

src/codeflare_sdk/__init__.py
snalawad: I meant to say that It should get updated with every release. We can remove it if its not required or makes it confusing.

ChristianZaccaria: Ohh I understand, that makes sense. I saw the comment as an "action item" so got confused. I'd suggest to remove the comment. Great work, thanks!

KPostOffice (commented):
Thanks for the contribution @snalawad !

src/codeflare_sdk/__init__.py
```diff
@@ -19,3 +19,13 @@
 from .job import RayJobClient
 
 from .utils import generate_cert
+
+from importlib.metadata import version, PackageNotFoundError
+
+try:
+    __version__ = version(
+        "codeflare-sdk"
```
KPostOffice, line 27: One other comment. I think:

```
__package__ or __name__
```
would also work here

snalawad: Won't this change the whole idea of getting the version details with codeflare_sdk.__version__ ??
If this is what needs to be used, I will make the change.

KPostOffice: Sorry, I think I wasn't quite clear. I meant:
`("codeflare-sdk")` -> `(__package__ or __name__)`, so just dynamically getting the package name rather than hard coding it.

KPostOffice: We can change it in the future if we need to for some reason. I'm happy to have this merged in. Thanks for the contribution @snalawad

KPostOffice (approved):

ChristianZaccaria (approved):
/lgtm thank you!!

PR #547, opened 2024-05-27
update SDK args

alexxfan:
# Issue link

[RHOAIENG-7200](https://issues.redhat.com/browse/RHOAIENG-7200)
# What changes have been made

What

Update the Cluster configuration in the SDK to make arguments clearer.

Currently the default config is as such:
```
cluster = Cluster(ClusterConfiguration(
    name='jobtest',
    namespace='default', # Update to your namespace
    num_workers=2,
    min_cpus=1,
    max_cpus=1,
    min_memory=4,
    max_memory=4,
    num_gpus=1,
    head_gpus=1,
    image="quay.io/project-codeflare/ray:latest-py39-cu118",
    write_to_file=False, # When enabled Ray Cluster yaml files are written to /HOME/.codeflare/resources 
    local_queue="local-queue-test" # Specify the local queue manually
)) 
```

How

Alter the following args:

* head_gpus -> num_head_gpus
* num_gpus ->  num_worker_gpus
* min_cpus -> worker_cpu_requests
* max_cpus -> worker_cpu_limits
* min_memory -> worker_memory_requests
* max_memory -> worker_memory_limits

# Verification steps
### Notebook server ODH/RHOAI/Local
* Clone this repository with `git clone https://github.com/project-codeflare/codeflare-sdk.git`
* Checkout this PR's branch
* Run `poetry build` - install if needed (`pip install poetry`)
* Run `pip install --force-reinstall dist/codeflare_sdk-0.0.0.dev0-py3-none-any.whl` 
* Restart your notebook kernel
## Manual Testing
* Open any of the guided demo notebooks.
* Follow step by step using the old arguments, warnings should be thrown when ClusterConfiguration is run.
* Rerun the notebook using the new, suggested arguments, no errors should be given.
## Checks
- [x] I've made sure the tests are passing. 
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] Testing is not required for this change

PR #579, opened 2024-07-02
Split head memory and cpu requests/limits

Bobbins228:
# Issue link

Closes: [RHOAIENG-9259](https://issues.redhat.com/browse/RHOAIENG-9259)
# What changes have been made

* Split the head cpu and memory resources to requests/limits similar to #547 
* Added depreciation warnings to the old vars `head_cpus` and `head_memory`
* Updated head/worker_extended_resource_request to include string values due to failing `get_cluster` method
* Updated notebook WF tests to reflect new parameters
* Updated existing e2e tests with new Parameters
* Added documentation for depreciating variables
# Verification steps

* Checkout this PR's branch
### Testing the depreciating args `head_cpus` and `head_memory`
Follow through the [basic Ray demo](https://github.com/project-codeflare/codeflare-sdk/blob/main/demo-notebooks/guided-demos/0_basic_ray.ipynb). Set the `head_cpus` and `head_memory` parameters to a value of your choosing.
You should get a warning that the parameters are being depreciated and to use the new ones.

The head cpu requests and limits should *both* equate the values you entered for the above.

### Testing the new requests/limits args
In the ClusterConfiguration add the parameters 
* head_cpu_requests
* head_cpu_limits
* head_memory_requests 
* head_memory_limits

Set them to values of your choosing and the head pod of the Ray Cluster should reflect these values.
## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


src/codeflare_sdk/cluster/config.py
```diff
@@ -36,8 +36,12 @@ class ClusterConfiguration:
     name: str
     namespace: str = None
     head_info: list = field(default_factory=list)
-    head_cpus: typing.Union[int, str] = 2
-    head_memory: typing.Union[int, str] = 8
+    head_cpu_requests: typing.Union[int, str] = 2
```
astefanutti, line 39: Would it be possible to check how the official Kubernetes Python client types resource requests and limits?

src/codeflare_sdk/cluster/model.py
```diff
@@ -72,8 +72,10 @@ class RayCluster:
...
     status: RayClusterStatus
-    head_cpus: int
-    head_mem: str
+    head_cpu_requests: int
+    head_cpu_limits: int
+    head_mem_requests: str
+    head_mem_limits: str
     head_gpu: int
     workers: int
     worker_mem_min: str
```
ChristianZaccaria, line 81: worker_mem_min: str
    worker_mem_max: str

Might be worth changing these to `worker_mem_requests/limits`

src/codeflare_sdk/cluster/config.py
Bobbins228: The official resource [request/limits](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1ResourceRequirements.md) is `dict(str, str)`

The SDK accepts both int and str and converts to str before setting the resources later in `generate_yaml.py`

astefanutti: What's the plan to support different kind of accelerators, like AMD GPUs or Intel Gaudi?

Bobbins228: Kevin has this PR up for supporting different accelerators #531 I think it is pretty much there we should try get it in before next release

ChristianZaccaria (commented):
There is one thing to note, perhaps unrelated to this PR, and is that a user can basically input ANY value in the ClusterConfiguration parameters. I.e., I can set to  head_cpu_requests=True, including head_gpus=True, and that is reflected in the yaml as a bool. I believe this is not the expected behaviour. - Note that this was tested on KinD as my OpenShift cluster isn't working at the moment.

Bobbins228:
@ChristianZaccaria 
This is not expected behaviour at all :(
I can have a look at adding some validation to ensure that the head/worker requests/limits are of the correct type.
Good catch!

ChristianZaccaria:
@Bobbins228 I couldn't get further, but I suppose maybe `cluster.up()` will already capture that and throw an error for using the wrong datatypes. However, you're right, there seems to be no validation when creating the yaml file.

Bobbins228:
@ChristianZaccaria This is insane! It seems you can pretty much set any of the variables to whatever type you like.
I will create a Jira for fixing the validation on all ClusterConfiguration parameters.

Bobbins228:
Applied do not merge label until [RHOAIENG-9259](https://issues.redhat.com/browse/RHOAIENG-9259) is a priority again.

docs/cluster-configuration.md
```diff
@@ -28,3 +30,17 @@ Note: 'quay.io/modh/ray:2.35.0-py39-cu121' is the default image used by the Code
 The `labels={"exampleLabel": "example"}` parameter can be used to apply additional labels to the RayCluster resource.
 
 After creating their `cluster`, a user can call `cluster.up()` and `cluster.down()` to respectively create or remove the Ray Cluster.
+
+
+## Depreciating Parameters
```
KPostOffice, line 35:
```suggestion
## Deprecating Parameters
```

docs/cluster-configuration.md
```diff
@@ -28,3 +30,17 @@ Note: 'quay.io/modh/ray:2.35.0-py39-cu121' is the default image used by the Code
...
+The following parameters of the `ClusterConfiguration` are being depreciated in release `v0.22.0`. <!-- TODO: When removing depreciated parameters update this statement -->
```
KPostOffice, line 36:
```suggestion
The following parameters of the `ClusterConfiguration` are being deprecated in release `v0.22.0`. <!-- TODO: When removing depreciated parameters update this statement -->
```

docs/cluster-configuration.md
```diff
@@ -28,3 +30,17 @@ Note: 'quay.io/modh/ray:2.35.0-py39-cu121' is the default image used by the Code
...
+| Depreciated Parameter | Replaced By |
```
KPostOffice, line 37:
```suggestion
| Deprecated Parameter | Replaced By |
```

KPostOffice (changes requested):
This looks good to me, just some docs changes

PR #580, opened 2024-07-02
Make Local Queue label optional

Bobbins228:
# Issue link

closes: [RHOAIENG-9348](https://issues.redhat.com/browse/RHOAIENG-9348)
# What changes have been made

Altered the local_queue functionality so if no default local queue exists or `local_queue` is not specified then the label will not be applied to the Ray Cluster.
# Verification steps

* Checkout this PR's branch
### Remove Kueue from your OpenShift Cluster
* In the RHOAI DSC set Kueue's management state to `Removed`
* Create a Ray Cluster in the `ClusterConfiguration` without a local queue specified.
* Ensure the Ray Cluster does not have a local queue label
### Kueue is enabled but no default Local Queue exists
* Ensure no default Local Queue exists in a given namespace
* Create a Ray Cluster without the local queue specified
* Ensure the created Ray Cluster does not have the local queue label.

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/utils/generate_yaml.py
+++ b/src/codeflare_sdk/utils/generate_yaml.py
@@ -30,6 +30,7 @@ from ..cluster.auth import api_config_handler, config_check
 from os import urandom
 from base64 import b64encode
 from urllib3.util import parse_url
+from kubernetes.client.exceptions import ApiException
 
 
 def read_template(template):
@@ -191,8 +192,11 @@ def get_default_kueue_name(namespace: str):
             namespace=namespace,
             plural="localqueues",
         )
-    except Exception as e:  # pragma: no cover
-        return _kube_api_error_handling(e)
+    except ApiException as e:  # pragma: no cover
+        if e.status == 404 or e.status == 403:
+            return
+        else:
+            return _kube_api_error_handling(e)
     for lq in local_queues["items"]:
         if (
             "annotations" in lq["metadata"]
@@ -201,9 +205,6 @@ def get_default_kueue_name(namespace: str):
             == "true"
         ):
             return lq["metadata"]["name"]
-    raise ValueError(
-        "Default Local Queue with kueue.x-k8s.io/default-queue: true annotation not found please create a default Local Queue or provide the local_queue name in Cluster Configuration"
-    )
 
 
 def local_queue_exists(namespace: str, local_queue_name: str):
@@ -228,7 +229,9 @@ def local_queue_exists(namespace: str, local_queue_name: str):
 
 def add_queue_label(item: dict, namespace: str, local_queue: Optional[str]):
     lq_name = local_queue or get_default_kueue_name(namespace)
-    if not local_queue_exists(namespace, lq_name):
+    if lq_name == None:
+        return
+    elif not local_queue_exists(namespace, lq_name):
         raise ValueError(
             "local_queue provided does not exist or is not in this namespace. Please provide the correct local_queue name in Cluster Configuration"
         )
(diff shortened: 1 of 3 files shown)


src/codeflare_sdk/cluster/cluster.py
```diff
@@ -152,6 +152,7 @@ def create_app_wrapper(self):
         write_to_file = self.config.write_to_file
         local_queue = self.config.local_queue
         labels = self.config.labels
+        kueue = self.config.kueue
```
astefanutti, line 155: Is the `kueue` parameter really necessary? I fail to see why `local_queue` is not enough. And if there is a "default" LocalQueue, then we can assume Kueue should be used and the label set?

Bobbins228: Perhaps I was too gun ho with this...
I will adjust the PR to match what you said above good catch Antonin

src/codeflare_sdk/utils/generate_yaml.py
```diff
@@ -191,8 +192,11 @@ def get_default_kueue_name(namespace: str):
             namespace=namespace,
             plural="localqueues",
         )
-    except Exception as e:  # pragma: no cover
-        return _kube_api_error_handling(e)
+    except ApiException as e:  # pragma: no cover
+        if e.status == 404 or e.status == 403:
```
astefanutti, line 196: What's there error code returned when the API is not there? is it 404 as well? I'm not sure about 403, WDYT?

Bobbins228: I'll test the "not found" possibility more thoroughly but as for the 403 forbidden error I came to that conclusion while testing the scenario where Kueue is set to `Removed` in the DSC.

The CRDs are not removed with the operator by design. What does change is the users permissions somehow. 

When trying to list the local queues in a namespace we are met with a 403 forbidden error.

astefanutti: Possibly the 403 can be explained because the RoleBinding that grant read permission on local queues  is managed as part of the Kueue component, so when it's removed, that RoleBinding is deleted.

Another scenario that'd be good to test is when the CRD is not present. It won't be possible to do it with RHOAI, given Kueue is installed by default. Maybe just trying on a cluster with only KubeRay installed.

ChristianZaccaria (commented):
Tested and works as expected:
- When not specifying LocalQueue in ClusterConfiguration but has default-queue annotation, LocalQueue label is added.
- When not specifying LocalQueue in ClusterConfiguration without the default-queue annotation, LocalQueue label is not added.
- When specifying the LocalQueue in ClusterConfiguration, LocalQueue label is added.

Thanks!

src/codeflare_sdk/utils/generate_yaml.py
Bobbins228: Okay after giving this a go it seems 404 is correct when kueue does not exist on the cluster and 403 is acceptable when kueue has been removed via a DSC

ChristianZaccaria (commented):
/lgtm just needs a rebase.

If you have a chance, please create a Jira for updating the notebooks after a release is a made. Thank you!

Bobbins228:
Rebased! 
@ChristianZaccaria good catch. I'll create a Jira for updating the nbs

Bobbins228:
@astefanutti @ChristianZaccaria I think the PR just requires a review approval in order to merge now :)

astefanutti (approved):

PR #593, opened 2024-07-16
Add Type validation for Cluster Configuration Parameters

Ygnas:
# Issue link

[https://issues.redhat.com/browse/RHOAIENG-9435](https://issues.redhat.com/browse/RHOAIENG-9435)
# What changes have been made

Added Type Validation for Parameters
# Verification steps

In Cluster Configuration change any of the parameters to a wrong type. Should get an error similar to this:
`'num_workers' should be of type <class 'int'>, got list`

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/config.py
+++ b/src/codeflare_sdk/cluster/config.py
@@ -18,10 +18,10 @@ which is used to specify resource requirements and other details when creating a
 Cluster object.
 """
 
-from dataclasses import dataclass, field
 import pathlib
-import typing
 import warnings
+from dataclasses import dataclass, field, fields
+from typing import Dict, List, Optional, Union, get_args, get_origin
 
 dir = pathlib.Path(__file__).parent.parent.resolve()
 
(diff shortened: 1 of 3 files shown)


Bobbins228 (commented):
Tested this out and works as expected /lgtm

src/codeflare_sdk/cluster/config.py
```diff
@@ -206,4 +203,24 @@ def _memory_to_resource(self):
...
+    def is_valid_type(value, expected_type):
+        """Check if the value matches the expected type."""
+        origin_type = get_origin(expected_type)
+        if origin_type is Union:
+            return any(isinstance(value, t) for t in get_args(expected_type))
+        if origin_type in {list, dict, tuple}:
+            return isinstance(value, origin_type)
+        return isinstance(value, expected_type)
+
+    local_queue: Optional[str] = None
```
KPostOffice, line 226: I'm unsure why this was so far down in the Class. Can you move it to be with the other instance variables?

src/codeflare_sdk/cluster/config.py
```diff
@@ -206,4 +203,24 @@ def _memory_to_resource(self):
...
+        for field_info in fields(self):
+            value = getattr(self, field_info.name)
+            expected_type = field_info.type
+            if not self.is_valid_type(value, expected_type):
+                raise TypeError(
+                    f"'{field_info.name}' should be of type {expected_type}, got {type(value).__name__}"
+                )
+
+    @staticmethod
+    def is_valid_type(value, expected_type):
```
KPostOffice, line 217: Small nit for function name
```suggestion
    def _is_type(value, expected_type):
```

src/codeflare_sdk/cluster/config.py
```diff
@@ -206,4 +203,24 @@ def _memory_to_resource(self):
             warnings.warn("max_memory is being deprecated, use worker_memory_limits")
             self.worker_memory_limits = f"{self.max_memory}G"
 
-    local_queue: str = None
+    def validate_types(self):
```
KPostOffice, line 206:
```suggestion
    def _validate_types(self):
```

src/codeflare_sdk/cluster/config.py
```diff
@@ -206,4 +203,24 @@ def _memory_to_resource(self):
...
+                )
+
+    @staticmethod
+    def is_valid_type(value, expected_type):
+        """Check if the value matches the expected type."""
+        origin_type = get_origin(expected_type)
+        if origin_type is Union:
+            return any(isinstance(value, t) for t in get_args(expected_type))
+        if origin_type in {list, dict, tuple}:
+            return isinstance(value, origin_type)
```
KPostOffice, line 223: Afaict this won't cover the case where the `dict`, `list`, or `tuple` are of the wrong type (i.e. List[str] won't throw an error for [1, 2 ,3]). Is this even possible? I'm more than happy to leave it as either future work or won't do if it is very complex

KPostOffice (commented):
Nice work! I have a few comments

src/codeflare_sdk/cluster/config.py
```diff
@@ -206,4 +206,34 @@ def _memory_to_resource(self):
...
 
-    local_queue: str = None
+    def _validate_types(self):
+        """Validate the types of all fields in the ClusterConfiguration dataclass."""
+        for field_info in fields(self):
+            value = getattr(self, field_info.name)
+            expected_type = field_info.type
+            if not self._is_type(value, expected_type):
+                raise TypeError(
+                    f"'{field_info.name}' should be of type {expected_type}, got {type(value)}"
```
KPostOffice, line 216:
```suggestion
                    f"'{field_info.name}' should be of type {expected_type}"
```
I think that this is probably verbose enough for users and removes confusion around containing classes (i.e. should be of type Dict[str, str] but got <class 'dict'>)

KPostOffice (commented):
This is great, everything works as expected, just a small comment on the TypeError content.

KPostOffice (approved):
/lgtm

PR #645, opened 2024-08-20
Add provision in odh-sync workflow to adjust Pipfile.cpu and Pipfile.…

abhijeet-dhumal:
# Issue link

[RHOAIENG-11525](https://issues.redhat.com/browse/RHOAIENG-11525)

# What changes have been made

Added provision in odh-sync workflow to adjust Pipfile.cpu and Pipfile.gpu with codeflare-sdk release version

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/odh-notebooks-sync.yml
+++ b/.github/workflows/odh-notebooks-sync.yml
@@ -53,6 +53,22 @@ jobs:
       - name: Update Pipfiles in accordance with Codeflare-SDK latest release
         run: |
           package_name=codeflare-sdk
+          install_package_using_pipenv(){
+            if ! pipenv install ${package_name}~="${CODEFLARE_RELEASE_VERSION}"; then
+              echo "Failed to install ${package_name} with version ${CODEFLARE_RELEASE_VERSION} in $dir"
+              exit 1
+            fi
+            # Lock dependencies, ensuring pre-release are included and clear previous state
+            if ! pipenv lock --pre --clear ; then
+              echo "Failed to lock dependencies"
+              exit 1
+            fi
+            # remove virtual env and clear cache
+            if ! pipenv --rm --clear ; then
+              echo "Failed to remove virtual environment"
+              exit 1
+            fi
+          }
           # Get the list of available versions for the package
           if ! versions=$(pipenv run pip-versions list $package_name);then
             echo "Failed to retrieve versions for $package_name"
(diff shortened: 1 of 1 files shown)


.github/workflows/odh-notebooks-sync.yml
```diff
@@ -95,24 +112,36 @@ jobs:
               cd "$dir"
               minimum_supported_python_version_major=$(echo "${MINIMUM_SUPPORTED_PYTHON_VERSION}" | awk -F '.' '{print $1}') #integer of MINIMUM_SUPPORTED_PYTHON_VERSION env variable
               minimum_supported_python_version_minor=$(echo "${MINIMUM_SUPPORTED_PYTHON_VERSION}" | awk -F '.' '{print $2}') #decimal of MINIMUM_SUPPORTED_PYTHON_VERSION env variable
-              pipfile_python_version=$(grep -E '^python_version' ./Pipfile | cut -d '"' -f 2) # extracted from pipfile
+              if ! [ -f "Pipfile" ]; then
+                if [ -f "Pipfile.cpu" ]; then
+                  pipfile_python_version=$(grep -E '^python_version' ./Pipfile.cpu | cut -d '"' -f 2) # extracted from pipfile.cpu
```
sutaakar, line 117: Do you need to extract version also from Pipfile.gpu?

abhijeet-dhumal: Actually it seems the[ script used](https://github.com/opendatahub-io/notebooks/blob/5573f26767fb41ea347a6339d01e113ebef35094/scripts/update_library_version.sh#L94) to update Pipfile.cpu and Pipfile.gpu are identical wrt python version and are being updated on same time.. so selecting pipfile version from one of them seems ok to me

jstourac: That is true - they should always match - basically the Python version should match the name of the directory.
We could probably put some comment about this here why we ignore the Pipfile.gpu value in this case, but otherwise I agree with this approach.

jstourac (approved):
/lgtm

Srihari1192 (approved):

sutaakar (approved):

Bobbins228 (changes requested):
@abhijeet-dhumal can you rebase this PR and we can proceed with approval

Bobbins228 (approved):
/approve
/lgtm

PR #658, opened 2024-09-06
[FIX] generate cert functionality

Bobbins228:
# Issue link

# What changes have been made

Added function `get_secret_name` Which uses a label selector to gather the secret name for the given Ray Cluster.
# Verification steps

* Checkout this PR's branch
## Testing
* Create a deployment of CodeFlare Operator from main 
* Run the basic interactive and local interactive demos 
* ensure that secrets are created when using `generate_cert()`
## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/utils/generate_cert.py
+++ b/src/codeflare_sdk/utils/generate_cert.py
@@ -21,6 +21,7 @@ from cryptography.x509.oid import NameOID
 import datetime
 from ..cluster.auth import config_check, api_config_handler
 from kubernetes import client, config
+from .kube_api_helpers import _kube_api_error_handling
 
 
 def generate_ca_cert(days: int = 30):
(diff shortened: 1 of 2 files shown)


src/codeflare_sdk/utils/generate_cert.py
```diff
@@ -74,6 +75,24 @@ def generate_ca_cert(days: int = 30):
...
 
+def get_secret_name(cluster_name, namespace, api_instance):
+    label_selector = f"ray.openshift.ai/cluster-name={cluster_name}"
+    try:
+        secrets = api_instance.list_namespaced_secret(
+            namespace, label_selector=label_selector
+        )
+        for secret in secrets.items:
+            if (
+                f"{cluster_name}-ca-secret-" in secret.metadata.name
```
sutaakar, line 86: Would it have sense to rather select secret based on `ray.openshift.ai/cluster-name` label?

astefanutti: It seems, but it's true there can still be multiple secrets for the same cluster?

sutaakar: Though, there may be more secrets with the label.

Bobbins228: The oauth config secret also shares this label so we need to be more specific

astefanutti: At a later point, we can add an annotation to the client Secret to tell them apart more robustly.

sutaakar: The label selector is already there, I overlooked it.

sutaakar (approved):

astefanutti (approved):

PR #665, opened 2024-09-10
Adjust CodeFlare SDK tests to run on FIPS

oksanabaza:
# Issue link

https://issues.redhat.com/browse/RHOAIENG-12655
# What changes have been made

Made minor fixes to the MNIST dataset download by customizing the MD5 check handling
# Verification steps

E2E tests should run successfully on the FIPS cluster
## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/tests/e2e/mnist.py
+++ b/tests/e2e/mnist.py
@@ -15,6 +15,7 @@
 import os
 
 import torch
+import torchvision.datasets.utils as utils
 import requests
 from pytorch_lightning import LightningModule, Trainer
 from pytorch_lightning.callbacks.progress import TQDMProgressBar
@@ -138,6 +139,11 @@ class LitMNIST(LightningModule):
     ####################
 
     def prepare_data(self):
+        def check_md5(fpath, md5=None):
+            return True
+
+        utils.check_md5 = check_md5
+
         # download
         print("Downloading MNIST dataset...")
 
--- a/tests/e2e/mnist_pip_requirements.txt
+++ b/tests/e2e/mnist_pip_requirements.txt
@@ -1,4 +1,4 @@
 pytorch_lightning==1.9.5
 torchmetrics==0.9.1
-torchvision==0.12.0
+torchvision==0.19.0
 minio


tests/e2e/utils_md5.py
```diff
@@ -0,0 +1,9 @@
+import torchvision.datasets.utils as utils
+
+
+def check_md5(fpath, md5=None):
+    return True
+
+
+def apply_md5_patch():
```
KPostOffice, line 8: relevant part of torchvision code can be found [here](https://github.com/pytorch/vision/blob/00e7fa164bfdfd302f0b471c18ce2fd6bd1a50bc/torchvision/datasets/utils.py#L35-L58)

This shouldn't be an issue running on py >= 3.9 afaict with the used for security flag set. I'm surprised it is necessary. Are we running e2e tests on 3.8 or something?

KPostOffice: it actually looks like this is due to running a very [old version](https://github.com/project-codeflare/codeflare-sdk/blob/0813065d3622493da35f1ba4755e0835a8377aed/tests/e2e/mnist_pip_requirements.txt#L3) of torchvision in e2e tests. 0.12 is from March of 2022. @ChristianZaccaria is there a reason we are using such an old version?

ChristianZaccaria: I guess they weren't updated at any stage before. We can try and bump the versions as done for the demo-notebooks [requirements.txt](https://github.com/project-codeflare/codeflare-sdk/blob/main/demo-notebooks/guided-demos/requirements.txt) file.

ChristianZaccaria: > Are we running e2e tests on 3.8 or something?

e2e tests are ran on python 3.9

KPostOffice: It is definitely the old torch vision version. @oksanabaza Can you please try upgrading this? (hopefully there aren't any api changes that will break the e2e tests)

oksanabaza: Thank you, I've updated the torch vision version from 0.12.0 to 0.19.0

abhijeet-dhumal (commented):
/lgtm 
Verified by running the mnist_ray_sdk_oauth test in FIPS enabled cluster with and without this PR update
Thanks @oksanabaza, Awesome work!

Before :
image (140)

After:
image (141)

Fiona-Waters (approved):
Was able to run successfully on a fips enabled cluster. Great work Oksana.
/approve

PR #666, opened 2024-09-10
Upgrade Ray to version 2.35

Fiona-Waters:
# Issue link

[RHOAIENG-12186](https://issues.redhat.com/browse/RHOAIENG-12186) 

# What changes have been made

All instances of the ray version and image have been updated to Ray 2.35 and `quay.io/modh/ray:2.35.0-py39-cu121`

# Verification steps

- Install RHOAI 2.14 nightly build (brew.registry.redhat.io/rh-osbs/iib:812075)
- Run the interactive demo notebooks from this PR and enure that they work as expected.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests


Fiona-Waters:
@sutaakar the e2e is failing with `pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock [--no-update]` to fix the lock file.` This command was run locally and updates were made to the poetry.lock file accordingly before creating this PR. 
Would it make sense to add `poetry lock --no-update` [here](https://github.com/project-codeflare/codeflare-sdk/blob/main/.github/workflows/e2e_tests.yaml#L123) in the action?

sutaakar:
ideally it would be good to track why there is a difference in `poetry.lock`

Fiona-Waters:
> @Fiona-Waters hmm, do you have any idea why is it failing?



> @Fiona-Waters hmm, do you have any idea why is it failing?

I ran poetry lock [--no-update] locally again, all that changed was the content-hash but that seems to have done the job. Thanks Karel :)

docs/cluster-configuration.md
```diff
@@ -23,7 +23,7 @@ cluster = Cluster(ClusterConfiguration(
     labels={"exampleLabel": "example", "secondLabel": "example"},
 ))
 ```
-Note: 'quay.io/rhoai/ray:2.23.0-py39-cu121' is the default community image used by the CodeFlare SDK for creating a RayCluster resource. If you have your own Ray image which suits your purposes, specify it in image field to override the default image.
```
KPostOffice, line 26: Can we add a comment about available ROCM images for AMD GPUs too? that way users can easily find them?

KPostOffice (commented):
Should we call out the new images which support AMD GPUs in our docs?

docs/cluster-configuration.md
Fiona-Waters: I've updated this, please let me know if its sufficient. These 2 images are available from the quay.io/modh org, the other torch ones are [here](https://quay.io/repository/rhoai/ray?tab=tags), should I add reference to those also?

KPostOffice: I think linking would be good. Something like
"A list of available images can be found here..."

Fiona-Waters: @astefanutti Do you think it is okay to link to these [images](https://quay.io/repository/rhoai/ray?tab=tags) here? Or we should wait until we have documented them with an explanation as to their compatibility etc?

astefanutti: @Fiona-Waters I think that's OK to mention the `quay.io/modh/ray:2.23.0-py39-rocm61` image as @KPostOffice suggested. We can state AMD accelerators support is dev preview at the moment if you think that's important to mention in upstream documentation.

I would not mention example images from `quay.io/rhoai`. Alternatively, we could point to the documentation on how to build custom images, linking to https://github.com/opendatahub-io/distributed-workloads/blob/main/images/runtime/examples/README.md.

Fiona-Waters: Ok will link to the modh rocm image and to that documentation. Thank you!

Fiona-Waters: @KPostOffice if you are happy with this we can resolve this conversation and merge. Thank you!

PR #671, opened 2024-09-16
remove repeated conditional branch logic

KPostOffice:
I noticed that this section of the code was a little repetitive while I was making some other changes. This should be logically equivalent


Bobbins228 (changes requested):
Logically this looks sound. Can you run the pre-commit step and I can test this out.

Bobbins228 (approved):
Tested this out locally and from a notebook works like it did before. 
Thanks Kevin! 
/lgtm 
/approve

PR #675, opened 2024-09-18
move current namespace logic to config

KPostOffice:
# What changes have been made
I was running into some issues testing locally without specifying a namespace. I realized we could default early and make the error much clearer

# Verification steps
Use the codeflare_sdk in a jupyter nb locally and don't specify namespace in your cluster configuration

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/cluster/auth.py
+++ b/src/codeflare_sdk/cluster/auth.py
@@ -219,3 +219,25 @@ def api_config_handler() -> Optional[client.ApiClient]:
         return api_client
     else:
         return None
+
+
+def get_current_namespace():  # pragma: no cover
+    config_check()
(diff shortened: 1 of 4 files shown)


Bobbins228 (changes requested):
Hey Kevin this looks really good. Is it cool if you update this to work with the refactored project? 

Perhaps `get_current_namespace` should live in kube_api_helpers WDYT?

Bobbins228:
Perhaps kube_api_helpers doesnt work as there could be a circular import. 
What would you think of a utils file in the utils module for helper functions like this? 

I say this because as we add more modules e.g. KFT Training Client we would probably prefer common functions like this to live there? WDYT?

KPostOffice:
I'll take a look at finding a better place for this. A `util` file definitely makes more sense than `auth.py`

PR #683, opened 2024-09-24
Delete custom notebook image

jiripetrlik:
We stoped building this image. I think we can delete it completely.


ChristianZaccaria (changes requested):
This removal is awesome, thanks!

There are a couple more things which could be done in the scope of this PR:
- Remove [nightly-image-build.yaml](https://github.com/project-codeflare/codeflare-sdk/blob/main/.github/workflows/nightly-image-build.yaml) workflow.
- Update the [Manual Releases section](https://github.com/project-codeflare/codeflare-sdk/blob/main/README.md?plain=1#L43) in the readme. TBH, I think we should remove that whole section and rely solely on Automated Releases.

jiripetrlik:
@ChristianZaccaria 
Thank you for spotting this! Both points were removed.

I don't have any strong opinion whether to remove also whole readme for manual releases. But this PR is about custom notebook image. Therefore I think to remove Readme is for different PR.

ChristianZaccaria:
@jiripetrlik sounds good! The removal of the Manual Release section could be made in a separate PR.

Wondering why the e2e is failing here? - I think it's unrelated, maybe it has to do with the faulty ray image being used in the SDK. There will be a fix soon.

ChristianZaccaria (commented):
/lgtm

ChristianZaccaria (approved):
/lgtm /approve thanks!!

ChristianZaccaria (approved):
/approved

PR #690, opened 2024-09-27
remove installation of packages in nb sync

KPostOffice:
# What changes have been made
remove line doing CF-SDK installation just lock with new dependency

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/odh-notebooks-sync.yml
+++ b/.github/workflows/odh-notebooks-sync.yml
@@ -62,10 +62,8 @@ jobs:
               mv "${1}" Pipfile
               mv "${2}" Pipfile.lock
             fi
-            if ! pipenv install ${package_name}~="${CODEFLARE_RELEASE_VERSION}"; then
-              echo "Failed to install ${package_name} with version ${CODEFLARE_RELEASE_VERSION} in $dir"
-              exit 1
-            fi
+            # replace existing version of cf-sdk with new version in Pipfile
+            sed -i "s/codeflare-sdk = .*$/codeflare-sdk = \"~=$CODEFLARE_RELEASE_VERSION\"/g" Pipfile
             # Lock dependencies, ensuring pre-release are included and clear previous state
             if ! pipenv lock --pre --clear ; then
               echo "Failed to lock dependencies"


.github/workflows/odh-notebooks-sync.yml
```diff
@@ -62,10 +62,8 @@ jobs:
               mv $1 Pipfile
               mv $2 Pipfile.lock
             fi
-            if ! pipenv install ${package_name}~="${CODEFLARE_RELEASE_VERSION}"; then
-              echo "Failed to install ${package_name} with version ${CODEFLARE_RELEASE_VERSION} in $dir"
-              exit 1
-            fi
+            # replace existing version of cf-sdk with new version in Pipfile
+            echo "$(sed 's/codeflare-sdk = .*$/codeflare-sdk = \"~=$CODEFLARE_RELEASE_VERSION\"/g' Pipfile | envsubst)" > Pipfile
```
jstourac, line 66: Maybe I'm missing something, but why not to use simply the `sed -i ...` ? E.g.:
```suggestion
            sed --in-place 's/codeflare-sdk = .*$/codeflare-sdk = "~=${CODEFLARE_RELEASE_VERSION}"/g' Pipfile
```
:thinking:

KPostOffice: Thanks @jstourac, no reason not to. My sed foo is lacking

KPostOffice: Jk, this doesn't seem to work

jstourac (approved):
/lgtm

.github/workflows/odh-notebooks-sync.yml
jstourac: Right, sorry, I didn't realize :slightly_smiling_face: I put it here without any testing. Thanks for the change!

Bobbins228 (approved):
/approve

jstourac:
Just realized - without this, the Pipfile.lock isn't updated, is it? :thinking:  Didn't check whether there is some call elsewhere in the file now.

PR #693, opened 2024-09-30
Add codecov for unit test coverage

varshaprasad96:
# Issue link

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -29,3 +29,8 @@ jobs:
           coverage run -m --source=src pytest -v tests/unit_test.py
           coverage=$(coverage report -m | tail -1 | tail -c 4 | head -c 2)
           if (( $coverage < 90 )); then echo "Coverage failed at ${coverage}%"; exit 1; else echo "Coverage passed, ${coverage}%"; fi
+        continue-on-error: true
+      - name: Upload to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}


varshaprasad96:
@ChristianZaccaria Could we merge this to check if coverage reports are being uploaded? Looks like it should automatically search for coverage files in the repo and add it - so no more path configuration is needed.

Ygnas (approved):

ChristianZaccaria (approved):
Good for merge on my side
/lgtm
/approve

.github/workflows/unit-tests.yml
```diff
@@ -29,3 +29,7 @@ jobs:
           coverage run -m --source=src pytest -v tests/unit_test.py
           coverage=$(coverage report -m | tail -1 | tail -c 4 | head -c 2)
           if (( $coverage < 90 )); then echo "Coverage failed at ${coverage}%"; exit 1; else echo "Coverage passed, ${coverage}%"; fi
```
ChristianZaccaria, line 31: On second thought, I think the code coverage report won't be uploaded if the previous step results in less than 90% coverage, exiting the workflow.

ChristianZaccaria (changes requested):
One thought on this actually:

.github/workflows/unit-tests.yml
varshaprasad96: That's a good catch! It should work now.

varshaprasad96:
/hold

The action seems to be failing with rate limiting error mentioning that a new token is required.

KPostOffice (approved):

ChristianZaccaria (approved):
/lgtm

PR #695, opened 2024-10-01
Refactor Ray and AppWrapper modules

Bobbins228:
# Issue link

Closes: [RHOAIENG-13332](https://issues.redhat.com/browse/RHOAIENG-13332)
# What changes have been made

Refactored the Ray and Appwrapper modules to fit the design doc
# Verification steps

E2E and Unit tests should be sufficient
## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


src/codeflare_sdk.egg-info/PKG-INFO
```diff
@@ -1,4 +1,4 @@
 Metadata-Version: 2.1
-Name: codeflare-sdk
+Name: codeflare_sdk
```
Bobbins228, line 2: This snuck through. I will update this PR after the original is updated

varshaprasad96:
The split looks good overall! 👍

src/codeflare_sdk/ray/cluster/generate_yaml.py
ChristianZaccaria, line 1: Not sure if this is in the scope of this PR, but wondering what we could do about helper functions in the `generate_yaml.py` file, such as:
```
def is_openshift_cluster()
def is_kind_cluster()

**Review by ChristianZaccaria — COMMENTED** [review]
Ran through an imports diagnostic tool, looks great! - Just left one comment

File: src/codeflare_sdk/ray/cluster/generate_yaml.py
Comment (Bobbins228) [reply src/codeflare_sdk/ray/cluster/generate_yaml.py:1]: I can capture this when we bring back the refactor ray cluster creation work :+1:

**Review by ChristianZaccaria — COMMENTED** [review]
/lgtm

**Review by Ygnas — APPROVED** [review]

File: src/codeflare_sdk/ray/appwrapper/awload.py [hunk 2]
```diff
@@ -23,8 +23,8 @@
 import yaml
 
 from kubernetes import client
-from ..common import _kube_api_error_handling
-from ..common.kubernetes_cluster.auth import (
+from ...common import _kube_api_error_handling
```
KPostOffice, line 26: Should we move to using absolute imports? i.e.
```
from codeflare.sdk.common import _kube_api_config_handling
```

KPostOffice (commented):
Is anyone using codeflare-sdk for scripting @varshaprasad96 ? Do we need to maintain the current imports for a few releases?

varshaprasad96:
> Do we need to maintain the current imports for a few releases?

This is a breaking change, but we are going to continue to support the API functionality as is - so it should not be a requirement to maintain the import paths. Specifying in the changelog and mentioning it in the documentation should be enough.

KPostOffice (approved):

PR #681, opened 2024-09-23
Add UI table with Cluster specs and action buttons

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-12526 

# What changes have been made

image
- Added UI table that will display existing clusters with their specs, and handle user interactions which include:
  - Selecting a cluster.
  - Deleting the selected cluster.
  - Opening in a new tab the Ray Dashboard of the selected cluster.
  - Opening in a new tab the Ray Dashboard Job view of the selected cluster.
- Formatted table to merge req~lim resources into one column.
- Formatted status field to change text color and add an icon based on the cluster status.
- Added unit tests.
- Added outputs.
- Updated Ray model with `cpu requests` and `cpu limits`.

TODO: 
- Add `view_clusters` to guided-notebooks by default? - Given `Cluster Up` and `Cluster Down` widgets will appear by default, I think it's a good idea to immediately encourage the user to use (and be aware) of this feature. We don't need to trigger the buttons for `test-guided-notebooks` tests as it will be covered by `ui-notebooks-test` tests. Only the expected outputs would have to be updated.
- Enhance unit tests.
- Add to UI tests
- Implement `TODO` left in `widgets.py` file. [ :heavy_check_mark: ]

# Verification steps

Install the codeflare-sdk and import:
```
from codeflare_sdk import view_clusters
```
There are several items to test which include:
- Create several clusters.
- Select a cluster.
- Click on each button.
- Switch clusters and click on each button.
- Verify spec displayed.
- Verify output.
- Repeat with AppWrappers.

## Checks
- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] Testing is not required for this change

PR #697, opened 2024-10-02
Refactor ipywidgets to its own module and enhancements

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-13330 

# What changes have been made

- Refactor widgets to its own module
- Converted UI table widgets to its own class.
- Enhance widgets unit tests and increased coverage = 100%

# Verification steps

- e2e tests and UI notebooks tests should continue to pass.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ChristianZaccaria:
@KPostOffice In the latest commit here, I've converted each individual widget from the table to its own class. Let me know hwhat you think, and if it meets with expectations, or if you had something else in mind wrt https://github.com/project-codeflare/codeflare-sdk/pull/681#discussion_r1773640000 Thank you!

KPostOffice (commented):
Nice! I like it. Less args is nice. There's a failing unit test, but otherwise lgtm I think

tests/unit_test.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 IBM, Red Hat
+# Copyright 2024 IBM, Red Hat
```
Bobbins228, line 1: Not sure about this update. Doesn't this reflect creation date? (Shouldn't matter really we are getting rid of this file anyway)

Bobbins228 (approved):
Left a comment on changing the year of the copyright we can discuss some other time. 
Overall this looks good and I appreciate the comment changes :) 

/lgtm
/approve

tests/unit_test.py
```diff
@@ -19,6 +19,7 @@
 import os
 import re
 import uuid
+from io import StringIO
```
Ygnas, line 22:
```suggestion

```
This file appears to have several unused imports. Would we want to remove those?

tests/unit_test.py
ChristianZaccaria: Ohh so that's how it works, I reverted that now thanks!

tests/unit_test.py
ChristianZaccaria: Good catch thanks! - I went ahead and removed all unused imports and variables from that file, and re-organised them.

PR #710, opened 2024-10-15
add function for listing lqs by flavors

KPostOffice:
# Issue link
https://issues.redhat.com/browse/RHOAIENG-13429

# What changes have been made
This lists local queues available to a user as well as the flavors for that LQ if available

# Verification steps
make and deploy kueue with latest changes from https://github.com/opendatahub-io/kueue/pull/52

Create Flavors, LQs, and CQ

```
from codeflare_sdk.common.kueue import list_local_queues

list_local_queues(...)
```
will update with more details when cluster resumes

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/__init__.py
+++ b/src/codeflare_sdk/__init__.py
@@ -21,6 +21,10 @@ from .common import (
     KubeConfigFileAuthentication,
 )
 
+from .common.kueue import (
+    list_local_queues,
+)
+
 from .common.utils import generate_cert
 from .common.utils.demos import copy_demo_nbs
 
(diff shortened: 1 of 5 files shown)


Bobbins228 (commented):
@KPostOffice Can you provide some verification steps? I am going to take a stab at testing this and I want to make sure I am setting up my environment correctly.
Thanks

Ygnas (approved):
Works as intended.

Bobbins228 (changes requested):
Can we add a simple import for this function?

Bobbins228 (commented):
@KPostOffice Works perfectly couple of notes. 
1. Failed on my cluster with regular user possibly because I am using an independent Kueue deployment. (verified a regular user can see localqueues on another clean cluster)
2. Maybe as a follow on to enhance this functionality, we can for scenarios where a user searches for x amount of flavors and 1 doesn't exist we could return the list with the flavors that do exist and give a warning about the non existent ones WDYT?

If you think we have time to implement option 2 in time for release that would be cool but if not I can approve now and we will scope it out in another Jira.

Great work!

Bobbins228 (approved):
/lgtm /approve

PR #711, opened 2024-10-16
Intuit python version for compatible Ray image

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-14578 

# What changes have been made

The Ray image to be used by the head and worker nodes will be based on the python version in the system.

Currently, we support py3.9, and we will now support Ray with py3.11.

# Verification steps

1. Launch a jupyter notebook server with python3.9.
2. Run through the `0_basic_ray.ipynb` notebook with the `write_to_file` parameter set to `True`.
3. See the Ray image used in the head and worker group specs.
4. Repeat but with python3.11.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/src/codeflare_sdk/ray/cluster/generate_yaml.py
+++ b/src/codeflare_sdk/ray/cluster/generate_yaml.py
@@ -20,6 +20,7 @@ This sub-module exists primarily to be used internally by the Cluster object
 import json
 import sys
 import typing
+import warnings
 import yaml
 import os
 import uuid
@@ -34,7 +35,7 @@ import codeflare_sdk
 
 SUPPORTED_PYTHON_VERSIONS = {
     "3.9": "quay.io/modh/ray@sha256:0d715f92570a2997381b7cafc0e224cfa25323f18b9545acfd23bc2b71576d06",
-    "3.11": "quay.io/modh/ray:2.35.0-py311-cu121",
+    "3.11": "quay.io/modh/ray@sha256:db667df1bc437a7b0965e8031e905d3ab04b86390d764d120e05ea5a5c18d1b4",
 }
 
 
@@ -96,12 +97,11 @@ def update_image(spec, image):
     containers = spec.get("containers")
     if not image:
         python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
-        try:
-            if python_version in SUPPORTED_PYTHON_VERSIONS:
-                image = SUPPORTED_PYTHON_VERSIONS[python_version]
-        except Exception:  # pragma: no cover
-            print(
-                f"Python version '{python_version}' is not supported. Only {', '.join(SUPPORTED_PYTHON_VERSIONS.keys())} are supported."
+        if python_version in SUPPORTED_PYTHON_VERSIONS:
+            image = SUPPORTED_PYTHON_VERSIONS[python_version]
+        else:
+            warnings.warn(
+                f"No default Ray image defined for {python_version}. Please provide your own image or use one of the following python versions: {', '.join(SUPPORTED_PYTHON_VERSIONS.keys())}."
             )
     for container in containers:
         container["image"] = image
--- a/src/codeflare_sdk/ray/cluster/generate_yaml.py
+++ b/src/codeflare_sdk/ray/cluster/generate_yaml.py
@@ -18,6 +18,7 @@ This sub-module exists primarily to be used internally by the Cluster object
 """
 
 import json
+import sys
 import typing
 import yaml
 import os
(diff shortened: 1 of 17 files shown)


src/codeflare_sdk/ray/cluster/generate_yaml.py
```diff
@@ -88,9 +89,14 @@ def update_names(
 
 def update_image(spec, image):
     containers = spec.get("containers")
-    if image != "":
-        for container in containers:
-            container["image"] = image
+    if not image:
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        if python_version.startswith("3.9"):
```
Bobbins228, line 94: Can we have some error handling here for if a user (for some reason) uses Py 3.12 or something so that they are reminded only Py 3.xx is supported 
Otherwise this looks good I can test this out soon

Bobbins228 (changes requested):

src/codeflare_sdk/ray/cluster/generate_yaml.py
Bobbins228: Maybe have a const like `SUPPORTED_PYTHON_VERSIONS=["3.9", "3.11"]` which could be used to compare WDYT?

ChristianZaccaria: I've enhanced the function now by using constants, great idea!!

src/codeflare_sdk/ray/cluster/generate_yaml.py
```diff
@@ -88,9 +94,16 @@ def update_names(
 
 def update_image(spec, image):
     containers = spec.get("containers")
-    if image != "":
-        for container in containers:
-            container["image"] = image
+    if not image:
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        if python_version in SUPPORTED_PYTHON_VERSIONS:
```
Bobbins228, line 99: Can you surround this in a try and except?

Bobbins228 (changes requested):

src/codeflare_sdk/ray/cluster/generate_yaml.py
```diff
@@ -88,9 +94,15 @@ def update_names(
...
-    if image != "":
-        for container in containers:
-            container["image"] = image
+    if not image:
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        try:
+            if python_version in SUPPORTED_PYTHON_VERSIONS:
+                image = SUPPORTED_PYTHON_VERSIONS[python_version]
+        except Exception: # pragma: no cover
+            print(f"Python version '{python_version}' is not supported. Only {', '.join(SUPPORTED_PYTHON_VERSIONS.keys())} are supported.")
```
Bobbins228: Perhaps a warning in case the user knows what they are doing/has a custom image and still wants to proceed.
If we could be more concise and say "officially supported" that might help too.

Bobbins228: Also in the warning maybe a "If you would like to proceed please specify an image or use one of the supported vers"

Bobbins228 (changes requested):

src/codeflare_sdk/ray/cluster/generate_yaml.py
```diff
@@ -88,9 +94,17 @@ def update_names(
...
-        for container in containers:
-            container["image"] = image
+    if not image:
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        try:
+            if python_version in SUPPORTED_PYTHON_VERSIONS:
+                image = SUPPORTED_PYTHON_VERSIONS[python_version]
+        except Exception:  # pragma: no cover
+            print(
+                f"Python version '{python_version}' is not supported. Only {', '.join(SUPPORTED_PYTHON_VERSIONS.keys())} are supported."
```
KPostOffice, line 104: Lets be explicit about the issue here. i.e. "No default Ray image defined for {python version}. Please provide your own image or use one of the following python versions."

KPostOffice, line 104: I think we could do without this try except. The conditional is clear as it is with just an `if` `else`

Bobbins228 (commented):
Tested on py3.9 & 3.11 and it works as expected! 

/lgtm

KPostOffice (approved):

Bobbins228 (approved):

PR #715, opened 2024-10-18
Dependabot Enhancements

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-11457 

# What changes have been made

Note: in this repository I have enabled merge queues in the `main` branch + in a [dummy branch](https://github.com/project-codeflare/codeflare-sdk/tree/test-odh-notebooks-sync) for testing purposes.
- Ignore updating patch versions to focus on major and minor updates.
- Limit number of PRs opened.
- Add dependabot labeler workflow to add required labels to automatically add PRs to merge queues.
- Add merge_group condition on test workflows to be ran on merge queues.

# Verification steps

Merge queues:
1. I added changes from this PR to a [dummy branch](https://github.com/project-codeflare/codeflare-sdk/tree/test-odh-notebooks-sync).
2. I created Dummy PR 1: https://github.com/project-codeflare/codeflare-sdk/pull/716 
3. I created Dummy PR 2: https://github.com/project-codeflare/codeflare-sdk/pull/717 
4. See both PRs were added to the [merge queue](https://github.com/project-codeflare/codeflare-sdk/queue/test-odh-notebooks-sync), and both were successfully merged at the exact same time.

Dependabot:
1. [PRs generated](https://github.com/ChristianZaccaria/codeflare-sdk/pulls?q=is%3Aopen+is%3Apr) in my fork for:
- `pyproject.toml` and `poetry.lock` files.
- `requirements.txt` files.
- `package.json` and `yarn.lock` files.

Once this PR is merged, we can make use of the CodeFlare Machine token to verify the Dependabot-Labeler workflow adds the required labels to add the Dependabot PRs to the merge queue.

# Points for discussion:
Before we merge, I would like to discuss with the team if we should perhaps change the way we merge PRs. To significantly reduce costs of using expensive GitHub Runners such as the ones with GPUs ran in e2e tests, perhaps the new flow could be:

> If PR passes all tests including pre-commit, unit-tests, and code-coverage (low costs), then the PR is added to the merge queue where the e2e test will run with changes from this PR and the number of all other PRs in the merge queue.

> If e2e fails, the PR is removed from the merge queue, if passes, the PR is merged with all other PRs in the merge queue.

This would mean, e2e tests will not run on each opened PR, or on each push, significantly reducing the running costs.

**Note:** we could still allow to run e2e tests in a PR triggered by a `non-merge-queue` label perhaps.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,7 @@ updates:
         update-types: ["version-update:semver-patch"]
     open-pull-requests-limit: 4
     labels:
+      - "dependabot"
       - "test-guided-notebooks"
 
   # pip means poetry in this case, this keeps poetry.lock up to date with constraints in pyproject.toml.
@@ -27,6 +28,7 @@ updates:
         update-types: ["version-update:semver-patch"]
     open-pull-requests-limit: 1
     labels:
+      - "dependabot"
       - "test-guided-notebooks"
 
   # npm means yarn in this case, this keeps yarn.lock up to date with constraints in package.json.
@@ -39,4 +41,5 @@ updates:
         update-types: ["version-update:semver-patch"]
     open-pull-requests-limit: 1
     labels:
+      - "dependabot"
       - "test-ui-notebooks"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,18 +4,16 @@
 version: 2
 updates:
   # This is to update requirements.txt files in the guided-demos, and e2e directories.
-  # The group configuration option is used to group updates for consistency across related directories.
   - package-ecosystem: "pip"
     directories:
       - "**/demo-notebooks/guided-demos*"
       - "/tests/e2e"
     schedule:
-      interval: "weekly"
-    groups:
-      requirements.txt:
-        patterns:
-          - "*"
-    open-pull-requests-limit: 10
+      interval: "daily"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+    open-pull-requests-limit: 4
     labels:
       - "test-guided-notebooks"
 
@@ -23,7 +21,22 @@ updates:
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 10
+      interval: "daily"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+    open-pull-requests-limit: 1
     labels:
       - "test-guided-notebooks"
+
+  # npm means yarn in this case, this keeps yarn.lock up to date with constraints in package.json.
+  - package-ecosystem: "npm"
+    directory: "/ui-tests"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+    open-pull-requests-limit: 1
+    labels:
+      - "test-ui-notebooks"
--- a/.github/workflows/dependabot-labeler.yaml
+++ b/.github/workflows/dependabot-labeler.yaml
@@ -4,12 +4,22 @@ name: Dependabot Labeler
 
 on:
   pull_request:
+    types: [ labeled, synchronize, opened, reopened ]
 
 jobs:
     add-approve-lgtm-label:
-        if: github.actor == 'dependabot[bot]' && contains(github.event.pull_request.labels.*.name, 'dependabot')
+        if: ${{ github.actor == 'dependabot[bot]' && contains(github.event.pull_request.labels.*.name, 'dependabot') }}
         runs-on: ubuntu-latest
+
+        # Permission required to edit a PR
+        permissions:
+          pull-requests: write
+          issues: write
+
         steps:
+            - name: Checkout code
+              uses: actions/checkout@v4
+
             - name: Add approve and lgtm labels to Dependabot PR
               run: |
                   gh pr edit ${{ github.event.pull_request.number }} --add-label "lgtm" --add-label "approved"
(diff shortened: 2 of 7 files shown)


ChristianZaccaria:
Before we merge, I would like to discuss with the team if we should perhaps change the way we merge PRs. To significantly reduce costs of using expensive GitHub Runners such as the ones with GPUs ran in e2e tests, perhaps the new flow could be:

If PR passes all tests including pre-commit, unit-tests, and code-coverage (low costs), then the PR is added to the merge queue where the e2e test will run with changes from this PR and the number of all other PRs in the merge queue.

If e2e fails, the PR is removed from the merge queue, if passes, the PR is merged with all other PRs in the merge queue.

This would mean, e2e tests will not run on each opened PR, or on each push, significantly reducing the running costs.

ChristianZaccaria:
PR ready for merge. - One thing to test here, if an approver adds the lgtm and approved labels, will this PR be sent to the merge queue or merged directly. There 'may' be conflicting processes between openshift prow and github merge queues. However, the default process is to `Merge when ready`.

.github/workflows/dependabot-labeler.yaml
```diff
@@ -0,0 +1,17 @@
+# This workflow file adds the 'lgtm' and 'approved' labels to Dependabot PRs
+# This is done to ensure that the PRs that pass e2e are automatically merged/added to merge-queues by the CodeFlare bot
+name: Dependabot Labeler
```
KPostOffice, line 3: Can you use [this](https://github.com/actions/labeler)?

.github/dependabot.yml
```diff
@@ -4,26 +4,42 @@
...
   # pip means poetry in this case, this keeps poetry.lock up to date with constraints in pyproject.toml.
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 10
+      interval: "daily"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
```
KPostOffice, line 28: Why are we ignoring patch updates?

KPostOffice (commented):
Looks good! This should keeps PR tab much cleaner

.github/dependabot.yml
ChristianZaccaria: Good question, this is to avoid creating a larger number of PRs and use of resources for minor patches. A PR will still be created for `x.y` releases. In scenarios where we would get a critical/high CVEs and is resolved in a patch we could manually address them. If it does become an ongoing issue and requires regular manual updates we can definitely remove this parameter.

.github/workflows/dependabot-labeler.yaml
ChristianZaccaria: Good call, it looks like we can pass the `GITHUB_TOKEN` too. Will update this thanks!

ChristianZaccaria: I've had a closer look at this, I think it’s a great feature but I believe for this particular use case it won’t work as expected. After some digging into how the labeler action works, I see it can automatically add specific labels based on the files that have been changed. I think this is great, and we should look to implement this for notebook or ui changes for example. 

The issue in this case is that the CodeFlare Machine needs to add the `approved` and `lgtm` labels to add the PR to the merge queue, so it wouldn't be ideal to make it add these labels based on file changes as these could be performed by different actors such as another user.

KPostOffice (commented):
A quick question about the `e2e` tests being run on all PRs in the merge queue. If this does fail, how does GitHub decide which PRs to back out of the merge queue? Can we run `e2e` specifically once a test has been backed out of the queue?

Bobbins228:
> Once this PR is merged, we can make use of the CodeFlare Machine token to verify the Dependabot-Labeler workflow adds the required labels to add the Dependabot PRs to the merge queue.

@ChristianZaccaria you can use your own PAT as the CodeFlare Machine Account Token for testing on your fork.

ChristianZaccaria:
@KPostOffice The flow of GitHub Merge queues goes like this:
For this example assume all following PRs are ready to merge.
1. We add PR-1 to the merge queue -> a new temporary branch is automatically created with changes from PR-1 and target branch -> e2e is ran against it.
2. While the e2e is running on PR-1, we add PR-2 to the merge queue before PR-1 e2e test passes -> a new new temp branch is created with changes from PR-2, PR-1, and the target branch. -> e2e tests run here now.

A few scenarios can happen at this point:
- Tests passed on the temp branch with {PR-2 + PR-1 + target branch}, then both PRs get merged at the same time.
- In a scenario where PR-1 e2e failed but passed when running with {PR-2 + PR-1 + target branch}, GitHub merge queues merges both as it proved they pass the required tests together.
- Another scenario, PR-1 fails e2e, and {PR-1 + PR-2 + target branch} fail e2e, GitHub Merge Queues will drop PR-1 from the queue and have PR-2 retested. If PR-2 passes on its own, it gets merged. For PR-1, we can then run e2e individually in the PR by adding the `e2e` label.

Bobbins228:
You can merge the branch into your own fork to access the labeller workflow :+1:

ChristianZaccaria:
Here is a dummy PR used to test the dependabot labeler in my fork: https://github.com/ChristianZaccaria/codeflare-sdk/pull/182 - works as expected.

Bobbins228:
Should the guided demo nbs label have been added to that PR? @ChristianZaccaria

.github/dependabot.yml
```diff
@@ -4,26 +4,42 @@
...
+    directory: "/ui-tests"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+    open-pull-requests-limit: 1
+    labels:
+      - "dependabot"
+      - "test-ui-notebooks"
```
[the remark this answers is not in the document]
Bobbins228: Ah of course!

ChristianZaccaria:
There are 2 items that can be tested during merge:
- lgtm and approved labels automatically send the PR to merge queue.
- PR in merge queue runs through e2e test.

Item to test after merge:
- Check newly created Dependabot PRs. - Check if all labels are added, e2e test is ran in merge queue, and PR gets merged automatically.

PR #718, opened 2024-10-18
e2e test for heterogenous cluster

Ygnas:
# Issue link

[https://issues.redhat.com/browse/RHOAIENG-13430](https://issues.redhat.com/browse/RHOAIENG-13430)

# What changes have been made

Added new e2e test for heterogenous cluster

# Verification steps

[kueue](https://github.com/kubernetes-sigs/kueue) pull this and do:
`IMAGE_REGISTRY=quay.io/<username>/kueues make image-local-push deploy`

> Don't forget to change \<username\>

After building and installing new kueue test can be run by:

**Openshift**: `poetry run pytest -v -s ./tests/e2e -m 'openshift'`

**Local kind**: `poetry run pytest -v -s ./tests/e2e/heterogeneous_clusters_kind_test.py`

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/tests/e2e/heterogeneous_clusters_kind_test.py
+++ b/tests/e2e/heterogeneous_clusters_kind_test.py
@@ -1,0 +1,74 @@
+from time import sleep
+import time
+from codeflare_sdk import (
+    Cluster,
+    ClusterConfiguration,
+)
+
+from codeflare_sdk.common.kueue.kueue import list_local_queues
+
+import pytest
+
+from support import *
+
+
+@pytest.mark.skip(reason="Skipping heterogenous cluster kind test")
+@pytest.mark.kind
+class TestHeterogeneousClustersKind:
+    def setup_method(self):
+        initialize_kubernetes_client(self)
+
+    def teardown_method(self):
+        delete_namespace(self)
+        delete_kueue_resources(self)
+
+    @pytest.mark.nvidia_gpu
+    def test_heterogeneous_clusters(self):
+        create_namespace(self)
+        create_kueue_resources(self, 2)
+        self.run_heterogeneous_clusters()
+
+    def run_heterogeneous_clusters(
+        self, gpu_resource_name="nvidia.com/gpu", number_of_gpus=0
+    ):
+        for flavor in self.resource_flavors:
+            node_labels = (
+                get_flavor_spec(self, flavor).get("spec", {}).get("nodeLabels", {})
+            )
+            expected_nodes = get_nodes_by_label(self, node_labels)
+
+            print(f"Expected nodes: {expected_nodes}")
+            cluster_name = f"test-ray-cluster-li-{flavor[-5:]}"
+            queues = list_local_queues(namespace=self.namespace, flavors=[flavor])
+            queue_name = queues[0]["name"] if queues else None
+            print(f"Using flavor: {flavor}, Queue: {queue_name}")
+            cluster = Cluster(
+                ClusterConfiguration(
+                    name=cluster_name,
+                    namespace=self.namespace,
+                    num_workers=1,
+                    head_cpu_requests="500m",
+                    head_cpu_limits="500m",
+                    head_memory_requests=2,
+                    head_memory_limits=2,
+                    worker_cpu_requests="500m",
+                    worker_cpu_limits=1,
+                    worker_memory_requests=1,
+                    worker_memory_limits=4,
+                    worker_extended_resource_requests={
+                        gpu_resource_name: number_of_gpus
+                    },
+                    write_to_file=True,
+                    verify_tls=False,
+                    local_queue=queue_name,
+                )
+            )
+            cluster.up()
+            sleep(5)
+            node_name = get_pod_node(self, self.namespace, cluster_name)
+            print(f"Cluster {cluster_name}-{flavor} is running on node: {node_name}")
+            sleep(5)
+            assert (
+                node_name in expected_nodes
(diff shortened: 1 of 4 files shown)


Srihari1192:
@Ygnas we may need e2e test for heterogenous cluster to support for openshift clusters aswell to run on downstream testing

tests/e2e/support.py
```diff
@@ -65,19 +65,30 @@ def create_namespace(self):
         return RuntimeError(e)
 
 
-def create_new_resource_flavor(self):
-    self.resource_flavor = f"test-resource-flavor-{random_choice()}"
-    create_resource_flavor(self, self.resource_flavor)
+def create_new_resource_flavor(self, num_flavors=2):
```
Srihari1192, line 68: @Ygnas Wondering does multiple resource flavours , cluster queues and local queues creation is required for all the tests or this scenario should be covered particularly for heterogeneous test, because this change is impacting for all the sdk tests of kueue setup

tests/e2e/heterogeneous_clusters_oauth_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+from codeflare_sdk.common.kueue.kueue import list_local_queues
+
+import pytest
+
+from support import *
+
+
+@pytest.mark.openshift
+class TestRayLocalInteractiveOauth:
```
Srihari1192, line 17: Can you rename the test name specific to heterogenous clusters

tests/e2e/heterogeneous_clusters_kind_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+from codeflare_sdk.common.kueue.kueue import list_local_queues
+
+import pytest
+
+from support import *
+
+
+@pytest.mark.kind
+class TestRayLocalInteractiveOauth:
```
Srihari1192, line 16: Can you rename the test name specific to heterogenous clusters

Srihari1192 (changes requested):

tests/e2e/support.py
Ygnas: You are correct these would not be required for the rest of the tests, will update

KPostOffice (approved):

tests/e2e/heterogeneous_clusters_kind_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+from codeflare_sdk.common.kueue.kueue import list_local_queues
+
+import pytest
+
+from support import *
+
+
+@pytest.mark.skip(reason="Skipping heterogenous cluster kind test")
+class TestHeterogenousClustersOauth:
```
Srihari1192, line 16:
```suggestion
@pytest.mark.skip(reason="Skipping heterogeneous cluster kind test")
@pytest.mark.kind
class TestHeterogeneousClustersKind:
```

tests/e2e/heterogeneous_clusters_oauth_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+from codeflare_sdk.common.kueue.kueue import list_local_queues
+
+import pytest
+
+from support import *
+
+
+@pytest.mark.openshift
+class TestHeterogenousClustersOauth:
```
Srihari1192, line 17:
```suggestion
class TestHeterogeneousClustersOauth:
```

tests/e2e/heterogeneous_clusters_oauth_test.py
```diff
@@ -0,0 +1,72 @@
...
+                    worker_cpu_limits=1,
+                    worker_memory_requests=1,
+                    worker_memory_limits=4,
+                    image=ray_image,
+                    verify_tls=False,
+                    local_queue=queue_name,
+                )
+            )
+            cluster.up()
+            time.sleep(5)
```
Srihari1192, line 64: Is there any possibilities to add wait for condition status rather than time.sleep

tests/e2e/heterogeneous_clusters_oauth_test.py
```diff
@@ -0,0 +1,72 @@
...
+            node_name = get_pod_node(self, self.namespace, cluster_name)
+            print(f"Cluster {cluster_name}-{flavor} is running on node: {node_name}")
+            time.sleep(5)
+            assert (
+                node_name not in used_nodes
+            ), f"Node {node_name} was already used by another flavor."
+            used_nodes.append(node_name)
```
Srihari1192, line 71: @Ygnas @KPostOffice Thinking Don't we need to assert cluster is successfully up or not  with different resource flavors?
For now we are just asserting pods getting assigned right node labels

Srihari1192 (changes requested):

tests/e2e/heterogeneous_clusters_oauth_test.py
KPostOffice: Successfully up might be good here, it would also be nice to have an "expected_nodes" for each iteration of this loop rather than using the `used_nodes` list. i.e. get a list of nodes that corresponds to the current flavor and make sure the pods end up on one of those nodes

tests/e2e/heterogeneous_clusters_oauth_test.py
KPostOffice: We could probably just use cluster.wait_ready I think

KPostOffice (approved):

KPostOffice (approved):

PR #741, opened 2024-10-25
Add refresh button to widgets UI

ChristianZaccaria:
# Issue link

Jira: https://issues.redhat.com/browse/RHOAIENG-14057 

# What changes have been made

- Added a refresh button to the UI:

  When the button is clicked, the DataFrame is refreshed. Refreshing the list of RayClusters, the spec, and the status of the RayClusters in the DataFrame.
   
  Additionally, when the button is clicked, the button is disabled until the K8s API request for the list of RayClusters has returned. - This usually takes around 1 second.
- Removed redundant code. - Using `self.namespace` variable as opposed to retrieving the namespace from the dataframe.

- Updated Sphinx docs.

- Added UI e2e test case.

- Small refactor to separate concerns.
# Verification steps

1. Checkout this PR and run `poetry build` to generate a new wheel.
2. Install the wheel with `pip install`.
3. Run through the widgets notebook.
4. Create a RayCluster.
5. Run `view_clusters()`
6. Create a 2nd RayCluster.
7. Click the button to refresh the DataFrame.

We can also test to refresh the DataFrame while the RayCluster has the status `Failed`. Once ready we can refresh the DataFrame to show current status, eventually being `Ready`.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


Bobbins228:
@ChristianZaccaria I thought that we were also going to add polling as well as the refresh button?

Bobbins228:
@ChristianZaccaria nevermind we should stick to our initial vote of option 0

src/codeflare_sdk/common/widgets/widgets.py
```diff
@@ -95,6 +101,7 @@ def _initialize_callbacks(self):
         self.ray_dashboard_button.on_click(
             lambda b: self._on_ray_dashboard_button_click(b)
         )
+        self.refresh_data_button.on_click(lambda b: self._on_refresh_button_click(b))
```
Ygnas, line 104:
```suggestion
        self.refresh_data_button.on_click(lambda b: self._on_refresh_data_button_click(b))
```

src/codeflare_sdk/common/widgets/widgets.py
```diff
@@ -171,15 +146,12 @@ def _on_list_jobs_button_click(self, b):
...
-        namespace = self.ray_clusters_df[
-            self.ray_clusters_df["Name"] == self.classification_widget.value
-        ]["Namespace"].values[0]
 
         # Suppress from Cluster Object initialisation widgets and outputs
         with widgets.Output(), contextlib.redirect_stdout(
             io.StringIO()
         ), contextlib.redirect_stderr(io.StringIO()):
-            cluster = Cluster(ClusterConfiguration(cluster_name, namespace))
+            cluster = Cluster(ClusterConfiguration(cluster_name, self.namespace))
```
Bobbins228, line 154: I have a small concern on this one iirc we have it set where you get a warning if you are using an unsupported Python Version so imagine if someone is using Py3.12 with their own 3.12 image there could potentially be some confusion for them. 

Even though we don't actively encourage users to use unsupported versions it still could be possible that they do for whatever their reasons. 
We could potentially bypass this warning all together by specifying a random string for the image. (It shouldn't matter here at all because they will never interact with this ClusterConfiguration) 

WDYT?

KPostOffice (approved):
/lgtm Tested and worked as expected

Bobbins228:
@ChristianZaccaria can you rebase this when you get the chance and we can approve again?

ChristianZaccaria:
Rebased

PR #778, opened 2024-12-09
Create ray cluster with ssa

akram:
# Issue link

# What changes have been made
Adding apply function.

# Verification steps

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [x] Testing is not required for this change

### DIFF
--- a/src/codeflare_sdk/ray/cluster/cluster.py
+++ b/src/codeflare_sdk/ray/cluster/cluster.py
@@ -52,8 +52,14 @@ import os
 import requests
 
 from kubernetes import config
+from kubernetes.dynamic import DynamicClient
+from kubernetes import client as k8s_client
+from kubernetes.client.rest import ApiException
+
 from kubernetes.client.rest import ApiException
 import warnings
+
+CF_SDK_FIELD_MANAGER = "codeflare-sdk"
 
 
 class Cluster:
@@ -84,6 +90,12 @@ class Cluster:
         if is_notebook():
             cluster_up_down_buttons(self)
 
+    def get_dynamic_client(self):  # pragma: no cover
+        return DynamicClient(get_api_client())
+
+    def config_check(self):
+        return config_check()
+
     @property
     def _client_headers(self):
         k8_client = get_api_client()
@@ -95,9 +107,7 @@ class Cluster:
 
     @property
     def _client_verify_tls(self):
-        if not _is_openshift_cluster or not self.config.verify_tls:
-            return False
-        return True
+        return _is_openshift_cluster and self.config.verify_tls
 
     @property
     def job_client(self):
@@ -121,7 +131,6 @@ class Cluster:
         Called upon cluster object creation, creates an AppWrapper yaml based on
         the specifications of the ClusterConfiguration.
         """
-
         if self.config.namespace is None:
             self.config.namespace = get_current_namespace()
             if self.config.namespace is None:
@@ -130,7 +139,6 @@ class Cluster:
                 raise TypeError(
                     f"Namespace {self.config.namespace} is of type {type(self.config.namespace)}. Check your Kubernetes Authentication."
                 )
-
         return build_ray_cluster(self)
 
     # creates a new cluster with the provided or default spec
@@ -139,10 +147,11 @@ class Cluster:
         Applies the Cluster yaml, pushing the resource request onto
         the Kueue localqueue.
         """
-
+        print(
+            "WARNING: The up() function is planned for deprecation in favor of apply()."
+        )
         # check if RayCluster CustomResourceDefinition exists if not throw RuntimeError
         self._throw_for_no_raycluster()
-
         namespace = self.config.namespace
 
         try:
@@ -176,6 +185,54 @@ class Cluster:
         except Exception as e:  # pragma: no cover
             return _kube_api_error_handling(e)
 
+    # Applies a new cluster with the provided or default spec
+    def apply(self, force=False):
+        """
+        Applies the Cluster yaml using server-side apply.
+        If 'force' is set to True, conflicts will be forced.
+        """
+        # check if RayCluster CustomResourceDefinition exists if not throw RuntimeError
+        self._throw_for_no_raycluster()
+        namespace = self.config.namespace
+        name = self.config.name
+        try:
+            self.config_check()
+            api_instance = client.CustomObjectsApi(get_api_client())
+            crds = self.get_dynamic_client().resources
+            if self.config.appwrapper:
+                api_version = "workload.codeflare.dev/v1beta2"
+                api_instance = crds.get(api_version=api_version, kind="AppWrapper")
+                # defaulting body to resource_yaml
+                body = self.resource_yaml
+                if self.config.write_to_file:
+                    # if write_to_file is True, load the file from AppWrapper yaml and update body
+                    with open(self.resource_yaml) as f:
+                        aw = yaml.load(f, Loader=yaml.FullLoader)
+                    body = aw
+                api_instance.server_side_apply(
+                    field_manager=CF_SDK_FIELD_MANAGER,
+                    group="workload.codeflare.dev",
+                    version="v1beta2",
+                    namespace=namespace,
+                    plural="appwrappers",
+                    body=body,
+                    force_conflicts=force,
+                )
+                print(
+                    f"AppWrapper: '{name}' configuration has successfully been applied"
+                )
+            else:
+                api_version = "ray.io/v1"
+                api_instance = crds.get(api_version=api_version, kind="RayCluster")
+                self._component_resources_apply(
+                    namespace=namespace, api_instance=api_instance
+                )
+                print(f"Ray Cluster: '{name}' has successfully been applied")
+        except AttributeError as e:
+            raise RuntimeError(f"Failed to initialize DynamicClient: {e}")
+        except Exception as e:  # pragma: no cover
+            return _kube_api_error_handling(e)
+
     def _throw_for_no_raycluster(self):
         api_instance = client.CustomObjectsApi(get_api_client())
         try:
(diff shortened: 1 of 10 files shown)


src/codeflare_sdk/ray/cluster/cluster.py
```diff
@@ -176,6 +186,50 @@ def up(self):
...
+        namespace = self.config.namespace
+        # Ensure Kubernetes configuration is loaded
+        try:
+            self.config_check()
+            # Get the RayCluster custom resource definition
+            api = self.get_dynamic_client().resources.get(
+                api_version="ray.io/v1", kind="RayCluster"
+            )
+        except Exception as e:
+            raise RuntimeError("Failed to get RayCluster resource: " + str(e))
```
Bobbins228, line 205:
```suggestion
        self._throw_for_no_raycluster()
```

src/codeflare_sdk/ray/cluster/cluster.py
```diff
@@ -176,6 +186,50 @@ def up(self):
...
+
+        # Read the YAML file and parse it into a dictionary
+        try:
+            with open(self.resource_yaml, "r") as f:
+                resource_body = yaml.safe_load(f)
+        except FileNotFoundError:
+            raise RuntimeError(f"Resource YAML file '{self.resource_yaml}' not found.")
+        except yaml.YAMLError as e:
+            raise ValueError(f"Failed to parse resource YAML: {e}")
```
Bobbins228, line 214: This `apply()` function only seems to work with written resources.
Have you thought about adding in support for in-memory resources like in [up()](https://github.com/project-codeflare/codeflare-sdk/blob/main/src/codeflare_sdk/ray/cluster/cluster.py#L503-L508)

Bobbins228 (changes requested):

Bobbins228 (commented):
Can we add user documentation for this PR? 
https://github.com/project-codeflare/codeflare-sdk/blob/main/docs/sphinx/user-docs/ray-cluster-interaction.rst

src/codeflare_sdk/ray/cluster/cluster.py
```diff
@@ -176,6 +185,45 @@ def up(self):
...
+        self._throw_for_no_raycluster()
+        namespace = self.config.namespace
+        name = self.config.name
+        try:
+            self.config_check()
+            api_instance = client.CustomObjectsApi(get_api_client())
+            crds = self.get_dynamic_client().resources
+            if self.config.appwrapper:
+                api_version = "workload.codeflare.dev/v1beta2"
+                api_instance = crds.get(api_version, kind="AppWrapper")
```
Bobbins228, line 204:
```suggestion
                api_instance = crds.get(api_version=api_version, kind="AppWrapper")
```
This should prevent a `TypeError` when trying to create an AppWrapper via `apply()`

Bobbins228 (changes requested):
Began testing this again on both Kind and OpenShift. 
It seems that Appwrapper is not working with the new changes when creating a cluster via `apply()`.

```
TypeError: Discoverer.get() takes 1 positional argument but 2 were given
```

src/codeflare_sdk/ray/cluster/cluster.py
Bobbins228: Even with the update to the api_instance there are still errors when trying to apply AppWrappers.
```
Response: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"PatchOptions.meta.k8s.io \\"\\" is invalid: fieldManager: Required value: is required for apply patch","reason":"Invalid","details":{"group":"meta.k8s.io","kind":"PatchOptions","causes":[{"reason":"FieldValueRequired","message":"Required value: is required for apply patch","field":"fieldManager"}]},"code":422}\n'
```

src/codeflare_sdk/ray/cluster/cluster.py
```diff
@@ -744,6 +802,20 @@ def _create_resources(yamls, namespace: str, api_instance: client.CustomObjectsA
     )
 
 
+def _apply_resources(
+    yamls, namespace: str, api_instance: client.CustomObjectsApi, force=False
+):
+    api_instance.server_side_apply(
+        field_manager="cluster-manager",
```
KPostOffice, line 809: This should probably be `cf-sdk` or something similar and pulled out into a variable, used both here and for the AppWrapper

src/codeflare_sdk/ray/cluster/cluster.py
```diff
@@ -84,6 +88,14 @@ def __init__(self, config: ClusterConfiguration):
         if is_notebook():
             cluster_up_down_buttons(self)
 
+    def get_dynamic_client(self):  # pragma: no cover
+        """Return a dynamic client, optionally mocked in tests."""
+        return DynamicClient(get_api_client())
+
+    def config_check(self):
+        """Return a dynamic client, optionally mocked in tests."""
```
KPostOffice, line 96: Is this doc string intended here?

src/codeflare_sdk/ray/cluster/cluster.py
```diff
@@ -744,6 +802,20 @@ def _create_resources(yamls, namespace: str, api_instance: client.CustomObjectsA
     )
 
 
+def _apply_resources(
```
KPostOffice, line 805: Could you just use a [partial](https://www.geeksforgeeks.org/partial-functions-python/) instead of pulling this into it's own function?

src/codeflare_sdk/ray/cluster/cluster.py
akram: I added the field-manager field to the server_side_apply call (for AppWrapper). That should have fixed it.
I cannot reproduce it now, and I see the AppWrapper being properly created.

akram:
> Began testing this again on both Kind and OpenShift. It seems that Appwrapper is not working with the new changes when creating a cluster via `apply()`.
> 
> ```
> TypeError: Discoverer.get() takes 1 positional argument but 2 were given
> ```

I have encapsulated the exception and was not able to reproduce the exact issue here. 
Instead I was seeing an issue related to `appwrapper` not being enabled.

Bobbins228 (commented):
Tested this again and Appwrapper is working for initial apply but I ran into a couple of issues when trying to update the AppWrapper and apply my changes. 

I kept running into this error with regards to the field manager even with `force=True`:
```
A conflict occurred with the RayCluster resource.
Only one RayCluster with the same name is allowed. Please delete or rename the existing RayCluster before creating a new one with the desired name.
Response: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Apply failed with 1 conflict: conflict with \\"codeflare-operator\\" using workload.codeflare.dev/v1beta2: .spec.components","reason":"Conflict","details":{"causes":[{"reason":"FieldManagerConflict","message":"conflict with \\"codeflare-operator\\" using workload.codeflare.dev/v1beta2","field":".spec.components"}]},"code":409}\n'
```

akram:
> Tested this again and Appwrapper is working for initial apply but I ran into a couple of issues when trying to update the AppWrapper and apply my changes.
> 
> I kept running into this error with regards to the field manager even with `force=True`:
> 
> ```
> A conflict occurred with the RayCluster resource.
> Only one RayCluster with the same name is allowed. Please delete or rename the existing RayCluster before creating a new one with the desired name.
> Response: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Apply failed with 1 conflict: conflict with \\"codeflare-operator\\" using workload.codeflare.dev/v1beta2: .spec.components","reason":"Conflict","details":{"causes":[{"reason":"FieldManagerConflict","message":"conflict with \\"codeflare-operator\\" using workload.codeflare.dev/v1beta2","field":".spec.components"}]},"code":409}\n'
> ```


nice catch. I reproduced it. There is 2 issues actually. For appwrappers, I was not propagating the force=True .
But, on top of that, changing the `spec.components[0].template` is also blocked by webhooks.
I tried to update the mutating webhook the same way I did for raycluster oauth proxy field, but that's not working.
The validating webhook is currently enforcing the `raycluster.template` to be immutable.

Trying to edit an appwrapper manually with `oc edit` leads to the same error.
```
(codeflare-sdk-py3.9) akram@wadez-m1 codeflare-sdk % oc edit appwrappers.workload.codeflare.dev jobtest
error: appwrappers.workload.codeflare.dev "jobtest" could not be patched: admission webhook "vappwrapper.kb.io" denied the request: spec.components[0].template.raw: Forbidden: attempt to change immutable field
```

dgrove-oss:
AppWrappers enforce that spec.components[*].template is immutable after creation, because updates may invalidate the PodSet info that they've extracted and given to Kueue.   

I'm missing the bigger context of what you are trying to do, but you will not be able to mutate a RayCluster wrapped in an AppWrapper in place.

akram:
> AppWrappers enforce that spec.components[*].template is immutable after creation, because updates may invalidate the PodSet info that they've extracted and given to Kueue.
> 
> I'm missing the bigger context of what you are trying to do, but you will not be able to mutate a RayCluster wrapped in an AppWrapper in place.

hi @dgrove-oss ,
the goal of this PR is to introduce an `apply()` function that would avoid cases where users have to do down+up to update rayclusers and appwrappers. That's an idempotent alternative to the `up()` function that can also makes resource creation more straitghtforward.

src/codeflare_sdk/ray/cluster/cluster.py
```diff
@@ -744,6 +809,20 @@ def _create_resources(yamls, namespace: str, api_instance: client.CustomObjectsA
...
```
KPostOffice, line 812: Can you rename this to make it clear that it is specific to ray clusters

KPostOffice (commented):
Overall lgtm, just one nit

src/codeflare_sdk/ray/cluster/cluster.py
akram: done

akram:
@KPostOffice and @Bobbins228 do you think we can merge it?
right now, error states are already handled in `_kube_api_error_handling` and explicit error messages are provided.

Bobbins228 (approved):
/lgtm

PR #828, opened 2025-05-15
feat(RHOAIENG-25241): disable ray usage metrics by default

kryanbeane:
# Issue link

# What changes have been made

Made `RAY_USAGE_STATS_ENABLED` default to = 0, which == disabled. Added unit tests to ensure enabledment and disablement. Also did manual testing and verified that a deployed custom build image has a default env var of RAY_USAGE_STATS_ENABLED = 0 on both head and worker pods

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kryanbeane:
hey @szaher, I looked at the method you mentioned, to put `RAY_USAGE_STATS_ENABLED`  in the ray cluster resource, and want to get your thoughts on this: I checked back on the meeting notes (ty gemini) and we had mentioned we want to let customers enable sending usage stats if they want. And so my head is going "what's easier for user to enable this if they want to". If we enable in the way this PR is, they just need to edit the ClusterConfiguration spec with the below :
```
envs = {
    "RAY_USAGE_STATS_ENABLED": "1"
}
```

If we did it the way you mentioned, I don't thinkl they would be able to reenable it. Maybe I'm wrong tho, wdyt?

chipspeak (approved):

PR #874, opened 2025-08-01
e2e test

chipspeak:
Testing e2e failure for a missing `-y` in an install. Just attempting to verify that its occurring on main also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# CodeFlare SDK
+# CodeFlare SDK -> TESTING CI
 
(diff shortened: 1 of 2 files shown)


chipspeak:
it fails here to installing `dnsmasq`. I think we'll need to add a `-y` flag to [codeflare-common](https://github.com/project-codeflare/codeflare-common/blob/main/github-actions/kind/action.yml#L97) . Within my [RayJobs PR](https://github.com/project-codeflare/codeflare-sdk/blob/9a0114142d980e054655f179842e205eb2c79817/.github/workflows/e2e_tests.yaml) I've copied the contents of the external action and added the flag and it seems to be progressing past this failure now.

PR #880, opened 2025-08-12
RHOAIENG-26482: Add GCS Fault Tolerance, Disable Usage Stats, Rename the RayJob scope RayCluster Config

kryanbeane:
# Jira link
[RHOAIENG-26482](https://issues.redhat.com//browse/RHOAIENG-26482)

# What changes have been made
* Added GCS fault tolerance, verification & future changes pending on the bug fix. 
* Renamed `RayJobClusterConfig` to `ManagedClusterConfig` - made more sense to me as the Cluster is managed by RayJobs, and RayJob being in the name was confusing
* Disable usage stats and didn't add ability to overwrite it. We initially made it changeable by the user as we didn't want to add new fields the user couldn't change. With 3.0 we have the opportunity to just disable the usage stats.

# Verification steps
- There are no verification steps yet - testing and merging this is dependant on the GCS Fault Tolerance bug being fixed first so this PR will sit until then. 

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,15 @@
+coverage:
+  precision: 2
+  round: down
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 2.5%
+    patch:
+      default:
+        target: 85%
+        threshold: 2.5%
+
+ignore:
+  - "**/__init__.py"
--- a/src/codeflare_sdk/__init__.py
+++ b/src/codeflare_sdk/__init__.py
@@ -11,7 +11,7 @@ from .ray import (
     AppWrapperStatus,
     RayJobClient,
     RayJob,
-    RayJobClusterConfig,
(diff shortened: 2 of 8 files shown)


chipspeak (approved):
LGTM! My only thought (and it's purely devils advocate) is that I feel we may need to revisit the name of the RayJob's cluster config again in the future. It's a definite improvement but I feel the use of the use of `managed` might still raise cluster admin connotations rather than the 'managed by the RayJob' implication.

Again, the nittiest of nits. Just think it merits talking about!

kryanbeane:
Getting this merged to unblock other tickets. There is a bug ticket for the GCS Fault tolerance / ray head pod persistence ticket feature. We've added to fix the ray job GCS stuff as part of that ticket (should be the same fix)

PR #886, opened 2025-08-20
feat(RHOAIENG-29391): Store entrypoint scripts in configMaps

chipspeak:
# Issue link
[Jira](https://issues.redhat.com/browse/RHOAIENG-29391)

# What changes have been made
RayJobs now place the training script content into configMaps to allow the RayCluster pods access to their contents during execution. In current implementation, this works as intended for lifecycled clusters but for existing clusters, the pods will need to be restarted after RayJob submission for the configMaps to be reflected in the CR. 

The logic detects the presence of `.py` files in the `entrypoint`, checks local directories for matches (locally or in a notebook) and then extracts the contents for placement in configMaps. The logic also parses the contents of the file using regex to determine if local imports (other python files) are needed. If they are, it also adds these.

The below verification steps will cover both use-cases but for the sake of brevity, attached screenshots will only demonstrate the results for the life-cycled cluster.

Before attempting this verification, you'll need to disable Kueue in your RHOAI installation.

# Verification steps
## Setup
1. Checkout this branch of the codeflare-sdk and build a `whl` file using poetry.
2. Drag this `whl` file into your notebook's directory if running in RHOAI or the top level of your local directory if running in an IDE.
3. Copy both of the files in [this gist](https://gist.github.com/chipspeak/1508faf2b753a4a181757dba4a036a94) to your local directory or notebook directory depending on where you're testing this.
4. Create a jupyter notebook and paste in the below cell to install the SDK.
```python
%pip install codeflare_sdk-0.0.0.dev0-py3-none-any.whl --force-reinstall
```
5. Next, authenticate via the codeflare-sdk using the below:
```python
from codeflare_sdk import TokenAuthentication

auth = TokenAuthentication(
    token = "your_token",
    server = "your_server",
    skip_tls=False
)

auth.login()
```
## Life-cycled Cluster
1. First, we'll need to initialise our cluster using the `ManagedClusterConfig` object. Paste the below into a new cell and execute it.
```python
from codeflare_sdk import RayJob, ManagedClusterConfig

cluster_config = ManagedClusterConfig(
    num_workers=2,
    worker_cpu_requests=1,
    worker_cpu_limits=1,
    worker_memory_requests=4,
    worker_memory_limits=4,
    head_accelerators={'nvidia.com/gpu': 0},
    worker_accelerators={'nvidia.com/gpu': 0},
)
```
2. Next, we can initialise and submit the RayJob by executing by pasting the below into a new cell and executing it.
```python
rayjob = RayJob(
    job_name="rayjob",
    namespace="rhods-notebooks", # or whatever namespace your notebook is in
    cluster_config=cluster_config,
    entrypoint="python test.py",
)

rayjob.submit()
```
3. You should see output in the notebook reflecting successful parsing of files and placement into config map similar to below.
```
2025-08-21 07:37:21,297 Creating new cluster: rayjob-cluster
2025-08-21 07:37:21,298 Initialized RayJob: rayjob in namespace: rhods-notebooks
2025-08-21 07:37:21,299 Found local script: test.py -> will mount at /home/ray/scripts/test.py
```
4. You can check the status of the job by pasting the below into a new cell and executing it.
```
rayjob.status()
```
5. The job should execute successfully and clean up the cluster upon completion. You can verify the configMap content by checking it in `oc` via the below command:
```
oc get configmap rayjob-scripts -n rhods-notebooks -o yaml 
```
6. If you prefer to check it in the OpenShift console, navigate to `workloads/ConfigMaps/rayjob-scripts` and you should be able to view the configMap content similar to the below screenshot.
Screenshot 2025-08-21 at 08 22 13

NOTE: If attempting to do this multiple times, you'll need to manually delete the instance of the RayJob CR to prevent duplicate errors.

## Existing Cluster
1. First, we'll need to create and apply a RayCluster. You can do this by copying the below into a new cell and executing it.
```python
from codeflare_sdk import Cluster, ClusterConfiguration, RayJob

    name='existing-cluster',
    head_extended_resource_requests={'nvidia.com/gpu':0},
    worker_extended_resource_requests={'nvidia.com/gpu':0},
    num_workers=2,
    worker_cpu_requests=2,
    worker_cpu_limits=2,
    worker_memory_requests=2,
    worker_memory_limits=4,
    namespace='rhods-notebooks'
))

cluster.apply()
```
2. Before attempting anything else, ensure that this cluster is running by copying the below into a new cell and executing it. You can re-run the cell until the cluster is up.
```python
cluster.status()
```
3. Once the cluster is up, paste the below into a new cell and execute it.
```python
rayjob = RayJob(
    job_name="sdk-test-job",
    cluster_name="existing-cluster",
    namespace="rhods-notebooks",
    entrypoint="python test.py",
)

rayjob.submit()
```
4. This will fail as the ConfigMap won't be reflected in the RayCluster CR. We'll need to delete the head and worker nodes and this instance of the RayJob CR. You can do this manually in the OpenShift console. The cluster's nodes should be in the same namespace as your notebook. 
5. The RayJob instance can be found by navigating to `administration/CustomResourceDefinitions` and then searching for 'RayJob'. Click on the result and then the `Instances` tab to locate the RayJob instance for deletion.
6. Once its deleted, re-run the RayJob cell again. 
7. You can check the RayJob's status by clicking the newly generated RayJob submission in your namespace. It should have a name with the pattern `sdk-test-job-random_suffix`.
8. Once you've clicked on it, you can access the logs tab and you should see output similar to be below screenshot.
Screenshot 2025-08-21 at 08 04 12
9. Once you've verified the above, make sure you clean up your resources using the same steps for RayJobs and then the below content in a cell for RayClusters.
```python
cluster.down()
```

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/src/codeflare_sdk/ray/rayjobs/config.py
+++ b/src/codeflare_sdk/ray/rayjobs/config.py
@@ -559,6 +559,13 @@ class ManagedClusterConfig:
         logger.info(
             f"Added script volume '{configmap_name}' to cluster config: mount_path={mount_path}"
         )
+        
+    def validate_configmap_size(self, scripts: Dict[str, str]) -> None:
+        total_size = sum(len(content.encode("utf-8")) for content in scripts.values())
+        if total_size > 1024 * 1024:  # 1MB
+            raise ValueError(
+                f"ConfigMap size exceeds 1MB limit. Total size: {total_size} bytes"
+            )
 
     def build_script_configmap_spec(
         self, job_name: str, namespace: str, scripts: Dict[str, str]
--- a/src/codeflare_sdk/ray/rayjobs/config.py
+++ b/src/codeflare_sdk/ray/rayjobs/config.py
@@ -20,7 +20,7 @@ Cluster object.
 
 import pathlib
 from dataclasses import dataclass, field, fields
-from typing import Dict, List, Optional, Union, get_args, get_origin, Any
+from typing import Dict, List, Optional, Union, get_args, get_origin, Any, Tuple
 from kubernetes.client import (
     V1ConfigMapVolumeSource,
     V1KeyToPath,
@@ -515,3 +515,88 @@ class ManagedClusterConfig:
             }
 
         return gcs_ft_options
+
+    def add_script_volumes(
+        self, configmap_name: str, mount_path: str = "/home/ray/scripts"
+    ):
+        """
+        Add script volume and mount references to cluster configuration.
+
+        Args:
+            configmap_name: Name of the ConfigMap containing scripts
+            mount_path: Where to mount scripts in containers (default: /home/ray/scripts)
+        """
+        # Check if script volume already exists
+        volume_name = "ray-job-scripts"
+        existing_volume = next(
+            (v for v in self.volumes if getattr(v, "name", None) == volume_name), None
+        )
+        if existing_volume:
+            logger.debug(f"Script volume '{volume_name}' already exists, skipping...")
+            return
+
+        # Check if script mount already exists
+        existing_mount = next(
+            (m for m in self.volume_mounts if getattr(m, "name", None) == volume_name),
+            None,
+        )
+        if existing_mount:
+            logger.debug(
+                f"Script volume mount '{volume_name}' already exists, skipping..."
+            )
+            return
+
+        # Add script volume to cluster configuration
+        script_volume = V1Volume(
+            name=volume_name, config_map=V1ConfigMapVolumeSource(name=configmap_name)
+        )
+        self.volumes.append(script_volume)
+
+        # Add script volume mount to cluster configuration
+        script_mount = V1VolumeMount(name=volume_name, mount_path=mount_path)
+        self.volume_mounts.append(script_mount)
+
+        logger.info(
+            f"Added script volume '{configmap_name}' to cluster config: mount_path={mount_path}"
+        )
+
+    def build_script_configmap_spec(
+        self, job_name: str, namespace: str, scripts: Dict[str, str]
+    ) -> Dict[str, Any]:
+        """
+        Build ConfigMap specification for scripts
+
+        Args:
+            job_name: Name of the RayJob (used for ConfigMap naming)
+            namespace: Kubernetes namespace
+            scripts: Dictionary of script_name -> script_content
+
+        Returns:
+            Dict: ConfigMap specification ready for Kubernetes API
+        """
+        configmap_name = f"{job_name}-scripts"
+        return {
+            "apiVersion": "v1",
+            "kind": "ConfigMap",
+            "metadata": {"name": configmap_name, "namespace": namespace},
+            "data": scripts,
+        }
+
+    def build_script_volume_specs(
+        self, configmap_name: str, mount_path: str = "/home/ray/scripts"
+    ) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+        """
+        Build volume and mount specifications for scripts
+
+        Args:
+            configmap_name: Name of the ConfigMap containing scripts
+            mount_path: Where to mount scripts in containers
+
+        Returns:
+            Tuple of (volume_spec, mount_spec) as dictionaries
+        """
+        volume_spec = {"name": "ray-job-scripts", "configMap": {"name": configmap_name}}
+
+        mount_spec = {"name": "ray-job-scripts", "mountPath": mount_path}
+
+        return volume_spec, mount_spec
--- a/src/codeflare_sdk/ray/rayjobs/config.py
+++ b/src/codeflare_sdk/ray/rayjobs/config.py
@@ -559,7 +559,7 @@ class ManagedClusterConfig:
         logger.info(
             f"Added script volume '{configmap_name}' to cluster config: mount_path={mount_path}"
         )
-        
+
     def validate_configmap_size(self, scripts: Dict[str, str]) -> None:
         total_size = sum(len(content.encode("utf-8")) for content in scripts.values())
         if total_size > 1024 * 1024:  # 1MB
--- a/src/codeflare_sdk/ray/rayjobs/rayjob.py
+++ b/src/codeflare_sdk/ray/rayjobs/rayjob.py
@@ -455,7 +455,7 @@ class RayJob:
         """Handle script volumes for new clusters (uses ManagedClusterConfig)."""
         # Validate ConfigMap size before creation
         self._cluster_config.validate_configmap_size(scripts)
-        
+
         # Build ConfigMap spec using config.py
         configmap_spec = self._cluster_config.build_script_configmap_spec(
             job_name=self.name, namespace=self.namespace, scripts=scripts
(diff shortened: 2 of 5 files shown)


pawelpaszki:
verification steps work as expected! great work and kudos for adding tests! I'll let others review too

src/codeflare_sdk/ray/rayjobs/config.py
```diff
@@ -515,3 +515,88 @@ def _build_gcs_ft_options(self) -> Dict[str, Any]:
...
+        self.volumes.append(script_volume)
+
+        # Add script volume mount to cluster configuration
+        script_mount = V1VolumeMount(name=volume_name, mount_path=mount_path)
+        self.volume_mounts.append(script_mount)
+
+        logger.info(
+            f"Added script volume '{configmap_name}' to cluster config: mount_path={mount_path}"
+        )
+
```
kryanbeane, line 562: This is a bit of a nit / a just-in-case but ConfigMaps have a 1MB limit for the data so might be a good idea to throw in a small check to validate the size. I don't think users would run into this often unless with huge scripts but it's one of those cases that we aren't 100% of the nature of the data science use case so no harm adding it

```suggestion
def _validate_configmap_size(self, scripts: Dict[str, str]) -> None:
    total_size = sum(len(content.encode('utf-8')) for content in scripts.values())
    if total_size > 1024 * 1024:  # 1MB limit
        raise ValueError(...) 
```

src/codeflare_sdk/ray/rayjobs/rayjob.py
```diff
@@ -17,7 +17,12 @@
 """
 
 import logging
+import os
+import re
+import ast
 from typing import Dict, Any, Optional, Tuple
+from kubernetes import client
+from ...common.kubernetes_cluster.auth import get_api_client
 from python_client.kuberay_job_api import RayjobApi
```
kryanbeane, line 27:
```suggestion
from python_client.kuberay_cluster_api import RayClusterApi

```

kryanbeane: GitHub won't let me suggest where I want to for reasons :( On line [148-149](https://github.com/project-codeflare/codeflare-sdk/pull/886/files#diff-a8df90d07b5a68ba7075e6332fa1a2f5ca1cb861693a429075b69e6368081301R148-R149) could we instanciate the cluster api alongside the job api?

src/codeflare_sdk/ray/rayjobs/rayjob.py
```diff
@@ -282,3 +299,265 @@ def _map_to_codeflare_status(
...
+        # Get existing RayCluster
+        api_instance = client.CustomObjectsApi(get_api_client())
+        try:
+            ray_cluster = api_instance.get_namespaced_custom_object(
+                group="ray.io",
+                version="v1",
+                namespace=self.namespace,
+                plural="rayclusters",
+                name=self.cluster_name,
+            )
```
kryanbeane, line 504:
```suggestion
        try:
            ray_cluster = self._cluster_api_.get_ray_cluster(
                name=self.cluster_name,
                k8s_namespace=self.namespace
            )
```

src/codeflare_sdk/ray/rayjobs/test_rayjob.py
```diff
@@ -992,3 +993,729 @@ def test_build_ray_cluster_spec_with_gcs_ft(mocker):
...
+    mock_api_class = mocker.patch("codeflare_sdk.ray.rayjobs.rayjob.RayjobApi")
+
+    # Mock CustomObjectsApi with error
+    mock_custom_api = mocker.patch("kubernetes.client.CustomObjectsApi")
+    mock_api_instance = mocker.Mock()
+    mock_custom_api.return_value = mock_api_instance
+
+    from kubernetes.client import ApiException
+
+    mock_api_instance.get_namespaced_custom_object.side_effect = ApiException(
```
kryanbeane, line 1592: This will need to be changed based on the api change above. Can just mock the cluster api instead

src/codeflare_sdk/ray/rayjobs/test_rayjob.py
```diff
@@ -992,3 +993,729 @@ def test_build_ray_cluster_spec_with_gcs_ft(mocker):
...
+    mocker.patch("kubernetes.config.load_kube_config")
+    mock_api_class = mocker.patch("codeflare_sdk.ray.rayjobs.rayjob.RayjobApi")
+
+    # Mock CustomObjectsApi
+    mock_custom_api = mocker.patch("kubernetes.client.CustomObjectsApi")
+    mock_api_instance = mocker.Mock()
+    mock_custom_api.return_value = mock_api_instance
+
+    # Mock successful get but failed patch
+    mock_api_instance.get_namespaced_custom_object.return_value = {
```
kryanbeane, line 1625: Same re api mock

src/codeflare_sdk/ray/rayjobs/rayjob.py
```diff
@@ -282,3 +299,265 @@ def _map_to_codeflare_status(
...
+        # Update the RayCluster
+        try:
+            api_instance.patch_namespaced_custom_object(
+                group="ray.io",
+                version="v1",
+                namespace=self.namespace,
+                plural="rayclusters",
+                name=self.cluster_name,
+                body=ray_cluster,
+            )
```
kryanbeane, line 556:
```suggestion
            self._cluster_api_.patch_ray_cluster(
                name=self.cluster_name,
                ray_patch: ??? # unsure here as the kuberay client code here is funny, i will investigate and get back to you
                k8s_namespace=self.namespace,
            )
```

src/codeflare_sdk/ray/rayjobs/rayjob.py
```diff
@@ -282,3 +299,265 @@ def _map_to_codeflare_status(
...
+        Extract local Python script files from entrypoint command, plus their dependencies.
+
+        Returns:
+            Dict of {script_name: script_content} if local scripts found, None otherwise
+        """
+        if not self.entrypoint:
+            return None
+
+        scripts = {}
+        mount_path = "/home/ray/scripts"
```
kryanbeane, line 314: nit: could we make this a constant somewhere since it's referenced in a good few places and if we for some reason need / want to change it in the future, it'd be easier if it was a constant. Not necessarily in a constants file but at the top level of the class or something (whatever works)

kryanbeane (changes requested):
This is great, validated that it works as expected. Thanks for the detailed verification steps that was v helpful

Requesting some changes mostly with the custom resource API you added. The python client is essentially a wrapper for that API so would make sense to use the python client (cluster api one) instead of instanciating a new one in here. Hold off though and let me double check the Cluster API of the python client is functional. If it's not, we can just go with the way you have here for now and I'll LGTM it. 

So only actual changes I'm suggesting here for now is the config map size validation and moving "/home/ray/scripts" to a constant. I'll get back to you re the cluster API

Thanks!

src/codeflare_sdk/ray/rayjobs/rayjob.py
kryanbeane: on the above ^ this will work, just pass in the ray cluster object into `ray_patch` like you're passing it into `body`

PR #895, opened 2025-09-01
task(RHOAIENG-33045): Update Runtime image SHAs for PY312

chipspeak:
# Issue link
[Jira Link](https://issues.redhat.com/browse/RHOAIENG-33045)

# What changes have been made
Updated CodeFlare SDK runtime image defaults to include the new Python 3.12 image. This has also been updated in the demo notebooks where referenced.

Additionally, I've moved our supported versions logic to `constants.py` as I feel this is a more intuitively central location for this. Similarly, a new `utils.py` file extracts the correct runtime image version to pull in place of the less central existing logic.

# Verification steps
* Checkout this branch.
* Build the `whl` file by running `poetry build`.
* Open any demo notebook and set the kernel to a Python 3.12 version.
* Install this version of the SDK by running `pip install <path_to_your_whl> --force-reinstall`
* Restart the kernel.
* Create a Ray Cluster.
* Check the head pod spec for the image tag or log in to its terminal and run `python3 --version` to verify that it is using Python 3.12.9.
* Repeat the above with a kernel set to Python 3.11 to verify that the old image is also used as needed.

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -57,7 +57,7 @@ jobs:
       - name: Set up specific Python version
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: '3.12'
           cache: 'pip' # caching pip dependencies
 
       - name: Setup NVidia GPU environment for KinD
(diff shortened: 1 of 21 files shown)


chipspeak:
I've gotten the same error multiple times with the same 0.0000000000008% difference in the assertion. I've updated the value to match the Python 3.12 difference just to see if the tests pass etc. I don't think there'd be any harm in keeping this new value as we won't need to test Python 3.11 in this test going forward.

kryanbeane (approved):
great improvements to the image selection thanks Pat, lgtm

pawelpaszki (approved):

PR #896, opened 2025-09-01
RHOAIENG-27792: Add stop, resubmit, & delete. Auto tear down config maps. E2Es

kryanbeane:
# Issue link
[RHOAIENG-27792
](https://issues.redhat.com/browse/RHOAIENG-27792)

# What changes have been made
* Added job.stop(), job.resubmit(), and job.delete()
* Added logic to automatically tear down the training script config map when a Training job is deleted
* Updated E2E test

# Verification steps
* Follow [this guided notebook](https://github.com/kryanbeane/codeflare-sdk/blob/25853efb43813ca75be87f457de314ad03063b1a/demo-notebooks/guided-demos/5_submit_rayjob_cr.ipynb) to  submit a job and run the below:

* Stop / Suspend the job
```python
job.stop()
```
Verify manually that the job has been suspended (spec.suspend = true) and that the RayCluster was torn down

* Resubmit the Job and let it run to completion
```python
job.resubmit()
```

* Delete the Job CR
```python
job.delete()
```

## Checks
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/tests/e2e/rayjob/lifecycled_cluster_oauth_test.py
+++ b/tests/e2e/rayjob/lifecycled_cluster_oauth_test.py
@@ -7,7 +7,10 @@ sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 from support import *
 
 from codeflare_sdk import RayJob, ManagedClusterConfig
-from codeflare_sdk.ray.rayjobs.status import CodeflareRayJobStatus
+from codeflare_sdk.ray.rayjobs.status import (
+    CodeflareRayJobStatus,
+    RayJobDeploymentStatus,
+)
 import kubernetes.client.rest
 from python_client.kuberay_job_api import RayjobApi
 from python_client.kuberay_cluster_api import RayClusterApi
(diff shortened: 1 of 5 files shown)


tests/e2e/rayjob/lifecycled_cluster_oauth_test.py
```diff
@@ -58,113 +51,95 @@ def run_rayjob_with_lifecycled_cluster_oauth(self):
...
+    def verify_cluster_cleanup(self, rayjob: RayJob, timeout: int = 60):
+        """Verify that the cluster created by the RayJob has been cleaned up."""
         elapsed_time = 0
-        check_interval = 5  # Check every 5 seconds
+        check_interval = 5
+        cluster_api = RayClusterApi()
 
         while elapsed_time < timeout:
             try:
-                # Try to get the RayCluster resource
```
kryanbeane, line 146: @pawelpaszki I moved to using the ClusterAPi Python client where possible since that's wrapping the k8s api. just for future reference with these tests if we are interacting with job or cluster CRs we should just use the upstream clients

pawelpaszki:
I have verified the changes while running the notebook and executed the test locally (with increased timeout - as per my comment). it works good. Thanks!

pawelpaszki (approved):

PR #1077, opened 2026-05-07
test(RHOAIENG-57445): added E2E test for RayCluster Autoscaling

kryanbeane:
# Issue link

https://redhat.atlassian.net/browse/RHOAIENG-57445

# What changes have been made

Add E2E tests for Ray in-tree autoscaling (non-Kueue path):

- `tests/e2e/autoscaling_raycluster_sdk_kind_test.py` — KinD lifecycle test (scale up + scale down)
- `tests/e2e/autoscaling_raycluster_sdk_oauth_test.py` — OpenShift/OAuth lifecycle test (scale up + scale down)
- `tests/e2e/autoscaling_load.py` — Ray workload script that creates CPU-bound tasks to trigger autoscaling
- `tests/e2e/support.py` — Added `wait_for_worker_count()` and `run_autoscaling_load_in_head_pod()` helpers

Both tests creaate an autoscaling-enabled RayCluster without Kueue resources, verify scale-up under load, then verify scale-down after idle timeout.

# Verification steps

**KinD:**
```bash
kubectl config use-context kind-kind
cd tests/e2e
poetry run pytest -vv -s autoscaling_raycluster_sdk_kind_test.py -m kind
```

**OpenShift:**
```bash
oc login <cluster>
cd tests/e2e
poetry run pytest -vv -s autoscaling_raycluster_sdk_oauth_test.py -m openshift
```

Expected: cluster scales from `min_workers=1` to `>=2` under load, then back to `1` after idle timeout.

## Checks
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/tests/e2e/autoscaling_raycluster_sdk_oauth_test.py
+++ b/tests/e2e/autoscaling_raycluster_sdk_oauth_test.py
@@ -1,0 +1,63 @@
+from time import sleep
+
+import pytest
+
+from codeflare_sdk import Cluster, ClusterConfiguration
+
+from support import *
+
+
+@pytest.mark.openshift
+@pytest.mark.tier1
+class TestRayClusterAutoscalingSDKOauth:
+    def setup_method(self):
+        initialize_kubernetes_client(self)
+
+    def teardown_method(self):
+        if hasattr(self, "auth_instance"):
+            cleanup_authentication(self.auth_instance)
+        delete_namespace(self)
+
+    @pytest.mark.timeout(1800)
+    def test_autoscaling_scale_up_and_down_openshift_oauth(self):
+        self.setup_method()
+
+        create_namespace(self)
+
+        ray_image = get_ray_image()
+        resources = get_platform_appropriate_resources()
+        self.auth_instance = authenticate_for_tests()
+
+        cluster_name = f"autoscale-{random_choice()}"
+
+        cluster = Cluster(
+            ClusterConfiguration(
+                name=cluster_name,
+                namespace=self.namespace,
+                enable_autoscaling=True,
+                min_workers=1,
+                max_workers=4,
+                image=ray_image,
+                write_to_file=True,
+                verify_tls=False,
+                **resources,
+            )
+        )
+
+        cluster.apply()
+        wait_ready_with_stuck_detection(cluster, timeout=900, dashboard_check=False)
+
+        # Verify initial state: 1 worker (min_workers)
+        wait_for_worker_count(self, cluster_name, lambda n: n == 1, timeout_s=600)
+
+        # Trigger scale-up via load script in head pod
+        run_autoscaling_load_in_head_pod(self, cluster_name, tasks=4, sleep_s=180)
+
+        # Verify scale-up
+        wait_for_worker_count(self, cluster_name, lambda n: n >= 2, timeout_s=900)
+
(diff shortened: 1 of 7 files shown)


tests/e2e/support.py
```diff
@@ -344,6 +346,60 @@ def run_kubectl_command(args):
...
+    pods = self.api_instance.list_namespaced_pod(self.namespace, label_selector=label)
+    if not pods.items:
+        raise RuntimeError(f"No head pod found for cluster {cluster_name}")
+    head_pod = pods.items[0].metadata.name
+
+    subprocess.check_call(
+        [
+            "kubectl",
+            "cp",
+            "./tests/e2e/autoscaling_load.py",
```
pawelpaszki:
```suggestion
            os.path.join(os.path.dirname(os.path.abspath(__file__)), "autoscaling_load.py"),
```

tests/e2e/autoscaling_raycluster_sdk_oauth_test.py
```diff
@@ -0,0 +1,63 @@
...
+        wait_ready_with_stuck_detection(cluster, timeout=900, dashboard_check=False)
+
+        # Verify initial state: 1 worker (min_workers)
+        wait_for_worker_count(self, cluster_name, lambda n: n == 1, timeout_s=600)
+
+        # Trigger scale-up via load script in head pod
+        run_autoscaling_load_in_head_pod(self, cluster_name, tasks=4, sleep_s=180)
+
+        # Verify scale-up
+        wait_for_worker_count(self, cluster_name, lambda n: n >= 2, timeout_s=900)
```
pawelpaszki: from cursor:

```
run_autoscaling_load_in_head_pod uses subprocess.check_call for the kubectl exec, which blocks until ray.get(futures) returns inside the pod — i.e., until all 4 Ray tasks finish. With sleep_s=180 and 2 available CPUs (1 head + 1 worker), this takes ~360s. By the time control returns to the test and wait_for_worker_count(n >= 2) starts, the new autoscaled workers may have already scaled back down (since the work is done).

The load submission should be asynchronous (subprocess.Popen in the background) so the scale-up can be checked while the work is still in flight.
```

pawelpaszki:
@kryanbeane I am getting resources issues on a standard PSI cluster. one of the workers is stuck pending (for this step: `(autoscaler +2s) Resized to 4 CPUs, 0 GPUs.`) with the following info:

```
0/6 nodes are available: 1 node(s) were unschedulable, 2 Insufficient memory, 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }. no new claims to deallocate, preemption: 0/6 nodes are available: 2 No preemption victims found for incoming pod, 4 Preemption is not helpful for scheduling.
```

also from the race condition mentioned in another commit, the test got stuck on this:

```
(autoscaler +2s) Tip: use `ray status` to view detailed cluster status. To disable these messages, set RAY_SCHEDULER_EVENTS=0.
(autoscaler +2s) Adding 1 node(s) of type small-group-autoscale-iontx.
(autoscaler +2s) Resized to 4 CPUs, 0 GPUs.
(autoscaler +4m5s) Removing 1 nodes of type small-group-autoscale-iontx (idle).
(autoscaler +4m5s) Resized to 3 CPUs, 0 GPUs.
```

and eventually itmed out:

```
FAILED
```

chipspeak (approved):

PR #1087, opened 2026-05-13
RHOAIENG-62238: fix sdk tests (byoidc and non-byoidc)

pawelpaszki:
# Issue link

# What changes have been made
[RHOAIENG-62238](https://redhat.atlassian.net/browse/RHOAIENG-62238): fix sdk tests (byoidc and non-byoidc)

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/tests/upgrade/01_raycluster_sdk_upgrade_test.py
+++ b/tests/upgrade/01_raycluster_sdk_upgrade_test.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 import requests
 from time import sleep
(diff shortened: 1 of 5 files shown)


tests/upgrade/01_raycluster_sdk_upgrade_test.py
```diff
@@ -125,81 +126,48 @@ def setup_method(self):
 
     def _is_byoidc_cluster(self):
```
kryanbeane: only one comment. I notice this function is duplicated. DYT it could be moved to support.py so we can just have 1 copy of it?

kryanbeane (approved):

PR #1090, opened 2026-05-14
fix: RayJob submit now uses the API client set via set_api_client()

szaher:
The vendored RayjobApi and RayClusterApi were creating their own
  Kubernetes clients by calling config.load_kube_config() directly,
  ignoring the client provided through set_api_client(). This caused
  403 errors when authenticating with a custom token, as the request
  would fall back to the default kubeconfig or in-cluster service account.

  Override the vendored API instances after construction with ones that
  use the SDK's global API client, consistent with how the rest of the
  SDK (Kueue, Cluster) handles authentication.

# Issue link

# What changes have been made

# Verification steps

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kryanbeane (commented):
would you add some test coverage pretty please

kryanbeane (approved):

PR #1097, opened 2026-06-03
V0.31 pre upgrade clean

pawelpaszki:
# Issue link

[RHOAIENG-63111](https://redhat.atlassian.net/browse/RHOAIENG-63111) — RHOAI 2.25 pre-upgrade: enable vanilla `pre_upgrade` tests and integrate migration script.

Parent epic: [RHOAIENG-63109](https://redhat.atlassian.net/browse/RHOAIENG-63109) (Ray upgrade qualification 2.25 → 3.x). Program: [RHAISTRAT-1519](https://redhat.atlassian.net/browse/RHAISTRAT-1519).

Closes [RHOAIENG-63111](https://redhat.atlassian.net/browse/RHOAIENG-63111)

---

# What changes have been made

This PR adds the **2.25 pre-upgrade qualification path** for codeflare-sdk. Post-upgrade migration (2→3 normalize/recreate) remains on the **`v0.36`** line; this branch is intentionally **pre_upgrade only**.

**Base branch:** `v0.31` (`origin/v0.31`, release **0.31.5**) — includes upstream Ray **2.52.1** and **cryptography 46.0.5** (not the older `v0.31.1-test-fix` dependency line).

### Enable vanilla pre-upgrade tests

- **Container test image** (`images/tests/`): Dockerfile, `entrypoint.sh`, `run-tests.sh`, RBAC manifest for the test user.
  - Default marker: **`pre_upgrade`** (rejects `-m post_upgrade` on this image).
  - Cluster auth detection (htpasswd/LDAP vs BYOIDC) aligned with test credentials in the env file.
- **Upgrade tests** (`tests/upgrade/`):
  - `01_raycluster_sdk_upgrade_test.py` — seed `RayCluster` **`mnist`** in **`test-ns-rayupgrade`** with Kueue objects (namespace label, LocalQueue, etc.).
  - `02_dashboard_ui_upgrade_test.py` — Workload Metrics UI check (optional path when seed succeeds).
  - `conftest.py` — UI fixtures; session ordering hooks.
- **UI support** (`tests/ui/`): page objects and login flow for OpenShift OAuth / BYOIDC dashboard tests.
- **`tests/e2e/support.py`**: namespace creation (409 tolerate, Kueue managed label), shared helpers for upgrade tests.
- **`pyproject.toml` / `poetry.lock`**: `selenium`, `webdriver-manager`; pytest markers `pre_upgrade`, `post_upgrade`, `ui`.
[...]

PR #1098, opened 2026-06-03
2.x 3.x upgrade tests post upgrade

pawelpaszki:
https://redhat.atlassian.net/browse/RHOAIENG-63111

# What changes have been made
added idempotent migration script to be executed before original post-upgrade tests to migrate (if applicable) ray clusters from 2.x to 3.x

# Verification steps
- run pre-upgrade tests with this [branch](https://github.com/project-codeflare/codeflare-sdk/pull/1097), 
- upgrade RHOAI to 3.5EA2
- build an image with this branch
- run post_upgrade tests with the new image

## Checks
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/scripts/migration/ray_cluster_migration.py
+++ b/scripts/migration/ray_cluster_migration.py
@@ -31,11 +31,8 @@ REQUIREMENTS:
 
 USAGE EXAMPLES:
 
-    # Step 1: Backup - All clusters (you'll be prompted for backup directory)
+    # Step 1: Backup - All clusters
     python ray_cluster_migration.py pre-upgrade
-
-    # Step 1: Backup - With specific directory
-    python ray_cluster_migration.py pre-upgrade ./my-backup-dir
 
     # Step 1: Backup - Specific namespace
     python ray_cluster_migration.py pre-upgrade --namespace my-ns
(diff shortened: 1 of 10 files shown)


scripts/migration/ray_cluster_migration.py
```diff
@@ -266,6 +299,229 @@ def _suspend_cluster(
...
+        if PRE_UPGRADE_BACKUP_ANNOTATION not in annotations:
+            return
+        del annotations[PRE_UPGRADE_BACKUP_ANNOTATION]
+        api_instance.patch_namespaced_custom_object(
+            group="ray.io",
+            version="v1",
+            namespace=namespace,
+            plural="rayclusters",
+            name=name,
+            body={"metadata": {"annotations": annotations}},
```
kryanbeane: nit: This patches all annotations. Only thing I'd worry about here is if for some reason another annotation is added or removed after the get above, and then we patch and change the whole annotation dict and not just the one we want to remove. Very unlikely scenario so marked as a nit

```suggestion
            body={"metadata": {"annotations": {PRE_UPGRADE_BACKUP_ANNOTATION: None}}},
```

kryanbeane (approved):
only one nit. Happy to lgtm but will hold incase you think we should fix that one comment 

/hold

scripts/migration/ray_cluster_migration.py
pawelpaszki: I'll add it and hold the PRs until we have a call to go through it

kryanbeane (approved):

